PR #2, opened 2019-08-16
Add FreeBSD (+others?) support to smartmon

pyro3d:
Adds FreeBSD support:
1) Changes the direct binary execution to use a variable (binary found with whereis)
2) Allows ATACAM driver (SATA drives in FreeBSD)

Tested on Debian 9, as well as FreeBSD 12.


smartmon.sh
```diff
@@ -22,6 +22,9 @@ $1 ~ /^ *[0-9]+$/ && $2 ~ /^[a-zA-Z0-9_-]+$/ {
 SMARTCTLAWK
 )"
 
+# Find the smartctl binary, usually in /usr/sbin (Linux) or /usr/local/sbin (FreeBSD)
+SMARTCTL_BIN="$(whereis smartctl | awk '{print $2}')"
```
dswarbrick, line 26: Wouldn't this essentially just be the same as omitting the path prefix from the `smartctl` commands, and letting the shell's PATH find it?

dswarbrick:
Closing due to inactivity.

PR #5, opened 2019-08-20
ADD 3ware/tw_cli script

ntavares:
```
# Prometheus exporter for 3ware RAID controllers
#
# Half of it based on "Nagios Plugin for 3ware RAID" from "Hari Sekhon",
# Ref: http://github.com/harisekhon/nagios-plugins
# ... with additions for full info (-I) gathering
```

### DIFF
--- a/tw_cli.py
+++ b/tw_cli.py
@@ -1,0 +1,391 @@
+#!/usr/bin/env python
+#
+# Prometheus exporter for 3ware RAID controllers
+#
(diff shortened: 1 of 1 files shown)


dswarbrick:
This script is not Python 3 compatible. Shouldn't we be aiming to ensure that all scripts in this repo are Python 3, considering that there is less than a month before the official EOL of Python 2.x?

PR #6, opened 2019-08-21
Add dmslave_info textfile collector

ssinha-ionos:
Add textfile collector to count the number of lun devices associated
with device-mapper slaves and expose as Prometheus metrics.

Signed-off-by: Saket Sinha <saket.sinha@cloud.ionos.com>

### DIFF
--- a/dmslave_info.py
+++ b/dmslave_info.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+"""
+Script to count the number of lun devices associated with device-mapper
+slaves and expose a summary as Prometheus metrics.
+"""
+
+import os
+import sys
+import argparse
+from prometheus_client import CollectorRegistry, Gauge, write_to_textfile
+
+sysfs = "/sys/devices/"
+
+def main(arguments):
+
+    parser = argparse.ArgumentParser(description="detect lun devices associated with DM",
+                                 epilog=__doc__)
+    parser.add_argument("-f", "--prom-file",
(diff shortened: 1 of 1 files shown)


dmslave_info.py
```diff
@@ -0,0 +1,45 @@
...
+    registry = CollectorRegistry()
+
+    gauge_dm_info = Gauge('node_dmslave_info',
+                          'Devicemapper slave information',
+                          ['dm_device', 'lun_name'],
+                          registry=registry)
+
+    dm = [x for x in os.listdir(sysfs+'virtual/block/') if x.startswith('dm')]
+    for dx in dm:
+        dm_sd =  os.listdir(sysfs+'virtual/block/'+dx+'/slaves/')
```
dswarbrick: `os.path.join()` would be a more Pythonic way of doing this.

dmslave_info.py
```diff
@@ -0,0 +1,45 @@
...
+    args = parser.parse_args(arguments)
+
+    registry = CollectorRegistry()
+
+    gauge_dm_info = Gauge('node_dmslave_info',
+                          'Devicemapper slave information',
+                          ['dm_device', 'lun_name'],
+                          registry=registry)
+
+    dm = [x for x in os.listdir(sysfs+'virtual/block/') if x.startswith('dm')]
```
dswarbrick: Why not just use `glob.glob(sysfs+'virtual/block/dm-*')` ?

dswarbrick:
Obsoleted by https://github.com/prometheus/procfs/pull/412

I think we can close this.

PR #10, opened 2019-09-26
Add Drive_Life_Remaining% and Fix parse error

imluyuan:
Add parsing special characters:
Can not parse special characters  when use Dell raid

### DIFF
--- a/smartmon.sh
+++ b/smartmon.sh
@@ -12,7 +12,7 @@
 
 parse_smartctl_attributes_awk="$(
   cat <<'SMARTCTLAWK'
-$1 ~ /^ *[0-9]+$/ && $2 ~ /^[a-zA-Z0-9_-]+$/ {
+$1 ~ /^ *[0-9]+$/ && $2 ~ /^[a-zA-Z0-9_-]+.*/ {
   gsub(/-/, "_");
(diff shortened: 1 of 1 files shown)


smartmon.sh
```diff
@@ -12,7 +12,7 @@
 
 parse_smartctl_attributes_awk="$(
   cat <<'SMARTCTLAWK'
-$1 ~ /^ *[0-9]+$/ && $2 ~ /^[a-zA-Z0-9_-]+$/ {
+$1 ~ /^ *[0-9]+$/ && $2 ~ /^[a-zA-Z0-9_-]+.*/ {
```
SuperQ: Let's not open this up to random character matching.

```suggestion
$1 ~ /^ *[0-9]+$/ && $2 ~ /^[a-zA-Z0-9_-]+%?$/ {
```

smartmon.sh
```diff
@@ -155,7 +156,7 @@ OUTPUTAWK
 
 format_output() {
   sort |
-    awk -F'{' "${output_format_awk}"
+    awk -F'{' "${output_format_awk} |sed -e 's/%//g"
```
SuperQ: Let's do this with `gsub(/%/, "");`, like we do for dash to underscore in the awk.

PR #15, opened 2019-11-14
update yum.sh to handle packages names with "-"

### LINKED ISSUE #14: yum.sh ignores any package with "-" in the name
The command `egrep '^\w+\.\w+'` in yum.sh excludes all packages with a "-" in the name, undercounting pending upgrades.

I'm not entirely sure what the best fix is at this point?

haraldkoch:
The egrep expression was ignoring packages with "-" in the name; this fixes #14.

### DIFF
--- a/yum.sh
+++ b/yum.sh
@@ -6,7 +6,7 @@
 # 
 # Based on apt.sh by Ben Kochie <superq@gmail.com>
 
-upgrades=$(/usr/bin/yum -q check-updates | awk 'BEGIN { mute=1 } /Obsoleting Packages/ { mute=0 } mute { print }' | egrep '^\w+\.\w+' | awk '{print $3}' | sort | uniq -c | awk '{print "yum_upgrades_pending{origin=\""$2"\"} "$1}')
+upgrades=$(/usr/bin/yum -q check-updates | awk 'BEGIN { mute=1 } /Obsoleting Packages/ { mute=0 } mute { print }' | egrep '^(\w+\-*)+\.\w+' | awk '{print $3}' | sort | uniq -c | awk '{print "yum_upgrades_pending{origin=\""$2"\"} "$1}')
 
 echo '# HELP yum_upgrades_pending Yum package pending updates by origin.'
 echo '# TYPE yum_upgrades_pending gauge'


yum.sh
```diff
@@ -6,7 +6,7 @@
 # 
 # Based on apt.sh by Ben Kochie <superq@gmail.com>
 
-upgrades=$(/usr/bin/yum -q check-updates | awk 'BEGIN { mute=1 } /Obsoleting Packages/ { mute=0 } mute { print }' | egrep '^\w+\.\w+' | awk '{print $3}' | sort | uniq -c | awk '{print "yum_upgrades_pending{origin=\""$2"\"} "$1}')
+upgrades=$(/usr/bin/yum -q check-updates | awk 'BEGIN { mute=1 } /Obsoleting Packages/ { mute=0 } mute { print }' | egrep '^(\w+\-*)+\.\w+' | awk '{print $3}' | sort | uniq -c | awk '{print "yum_upgrades_pending{origin=\""$2"\"} "$1}')
```
SuperQ: Maybe better to make this a match like `grep -P '[\w-]+'`. Notice that this is PCRE mode, not GNU grep regexp mode.

```suggestion
upgrades=$(/usr/bin/yum -q check-updates | awk 'BEGIN { mute=1 } /Obsoleting Packages/ { mute=0 } mute { print }' | grep -P '^[\w-]+\.\w+' | awk '{print $3}' | sort | uniq -c | awk '{print "yum_upgrades_pending{origin=\""$2"\"} "$1}')
```

candlerb: I confirm that `grep -P '^[\w-]+\.\w+'` works on CentOS 6. However it misses this one:

```
libstdc++.x86_64                      4.4.7-23.el6                       base
```

Obviously `grep -P '^[\w+-]+\.\w+\s'` would work.

I am not sure why that grep is there. The only thing I see which shouldn't be there is a blank line.  Maybe it could be something even more generic like:

```
egrep '^[[:print:]]+\.[[:print:]]+'
```

SuperQ: I don't have any systems with yum around, so I don't have any example output to recommend improvements.

But the `:print:` suggestion seems reasonable.

haraldkoch:
I'm off work until January, but I'll try the :print: variant when I get back, and update the pull request if it works?

yum.sh
candlerb: I have opened PR #33 which also simplifies the pipeline slightly (removes one level of egrep and awk)

haraldkoch:
Closed by #33

PR #21, opened 2019-12-06
Include drive serial number as label in megaraid_pd_info

candlerb:
Helpful for detecting whether a drive has been unplugged and put back in, or replaced with a different drive.

Signed-off-by: Brian Candler <b.candler@pobox.com>


RichiH:
Thanks

PR #22, opened 2019-12-06
Fetch detailed info for drives with no enclosure (/cX/sZ)

candlerb:
Without this patch, you get firmware/serial info for drives in
/cX/eY/sZ but not those in /cX/sZ

Signed-off-by: Brian Candler <b.candler@pobox.com>


dswarbrick:
Please resolve conflicts if you still wish this to be considered.

PR #26, opened 2019-12-09
Suppress smart_healthy metric if SMART not enabled

candlerb:
This makes it easier to distinguish "health unknown" from "health bad"

### DIFF
--- a/smartmon.sh
+++ b/smartmon.sh
@@ -102,7 +102,7 @@ parse_smartctl_scsi_attributes() {
 }
 
 parse_smartctl_info() {
-  local -i smart_available=0 smart_enabled=0 smart_healthy=0
+  local -i smart_available=0 smart_enabled=0 smart_healthy=
   local disk="$1" disk_type="$2"
   local model_family='' device_model='' serial_number='' fw_version='' vendor='' product='' revision='' lun_id=''
   while read line; do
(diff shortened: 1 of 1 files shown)


SuperQ:
We only do boolean values for these metrics, adding an additional set state isn't what we want to do here. It should be perfectly possible to get what you need by using PromQL expressions like `smart_available == 1 and smart_healthy == 1`.

candlerb:
How about suppressing smart_healthy and smart_enabled metrics entirely if SMART is not available?

SuperQ:
Yes, that would be a reasonable option. Then you can do things like `smart_available unless smart_healthy`.

candlerb:
Even documenting the metric semantics would be an improvement. I was surprised to find so many disks with apparently "not healthy" state, and had to dig into the source to work out why.

Something else that's surprising at the moment is that if SMART is enabled, you get smart_available=0:

```
    if [[ "${info_type}" == 'SMART_support_is' ]]; then
      case "${info_value:0:7}" in
      Enabled) smart_enabled=1 ;;
      Availab) smart_available=1 ;;
      Unavail) smart_available=0 ;;
      esac
    fi
```

Therefore the test you proposed (`smart_available == 1 and smart_healthy == 1`) would not work, although `smart_enabled == 1 and smart_healthy == 1` ought to.

What do you think about also setting both smart_available and smart_enabled to 1 when it is enabled?

(SMART enabled/available/unavailable might make more sense as a label, but that's a bigger change)

PR #28, opened 2019-12-12
Dedup attributes

0x5a17ed:
Pull request contains a few fixes:

* A previous change introduced a regression and cause a few SMART tests to fail. (namely device_is_active) This is fixed by allowing `smart_ctl` to fail again and opt out of error checks explicitly. 
* Small fix that caused python to interpret `\d` as en escape sequence.
* Add a `--wakeup-disks` command line argument to optionally wake up disks and query them for their SMART values.
* deduplicate SMART attributes when reporting them as metrics to prometheus.

The last change is the most controversial one I assume. I found a disk that reports the same attribute name under different IDs.

```
171 Program_Fail_Count      0x000a   000   000   000    Old_age   Always       -       7
181 Program_Fail_Count      0x000a   000   000   000    Old_age   Always       -       7
```

This causes the script to produce the output

```
smartmon_attr_value{name="program_fail_count",disk="/dev/pass1"} 0
smartmon_attr_value{name="program_fail_count",disk="/dev/pass1"} 0
```

which in turn causes prometheus to not be happy about it.

The fix causes the script to deduplicate these attribute names and only report them once. A different approach would probably have been to change the output altogether and prepend the attribute id in front of the name so that `program_fail_count` becomes `attr171_program_fail_count`. This approach would be backwards incompatible though.

Please let me know what you think.

### DIFF
--- a/smartmon.py
+++ b/smartmon.py
@@ -7,6 +7,7 @@ import decimal
 import re
 import shlex
 import subprocess
+import sys
 
 device_info_re = re.compile(r'^(?P<k>[^:]+?)(?:(?:\sis|):)\s*(?P<v>.*)$')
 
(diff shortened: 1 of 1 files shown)


edgarasg:
@SuperQ can you please review this? Thanks.

PR #29, opened 2019-12-13
Add zfs_zpool.sh script

candlerb:
Useful for monitoring zpool health, zvolumes, and zfs filesystems which are not mounted

Also zfs-specific stats which can't be seen from node_filesystem metrics (e.g. compressratio)


grahamc:
For whatever it is worth, I use this to track the age of the latest snapshot:

```
        zfs list -Hp -t snapshot -o name,creation \
          | sed -e 's#@.*\s# #' \
          | awk '
              {
                if (last[$1] < $2) {
                  last[$1]=$2
                }
              }
              END {
                for (m in last) {
                  printf "zfs_snapshot_age_seconds{dataset=\"%s\"} %s\n", m, last[m];
                }
              }
            ' \
          | sponge znapzend-snaps.prom
```

zfs_zpool.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash -eu
```
dswarbrick, line 1: We're trying to standardize on `/usr/bin/env ...` for shebangs in this repo.
```suggestion
#!/usr/bin/env bash

 set -eu
```

candlerb: Done (and rebased on master)

Issue #27, opened 2019-12-12
storcli.py shows battery_backup_healthy when it needs attention


candlerb:
I have some megaraid controllers which are returning the following:

```
megaraid_healthy 0   <== there's a problem
megaraid_failed 0
megaraid_degraded 0
megaraid_battery_backup_healthy 1
```

This is odd: the controller says it needs attention, but it's not obvious why.

On closer inspection: storcli.py returns `battery_backup_healthy 1` if the BBU state is 0 or 32.  I'm getting 32, and the battery is also "Degraded":

```
# /opt/MegaRAID/storcli/storcli64 /cALL show all J | less
...
                "Status" : {
                  ==>   "Controller Status" : "Needs Attention",
                        "Memory Correctable Errors" : 0,
                        "Memory Uncorrectable Errors" : 0,
                        "ECC Bucket Count" : 0,
                        "Any Offline VD Cache Preserved" : "No",
                  ==>   "BBU Status" : 32,
                        "PD Firmware Download in progress" : "No",
                        "Support PD Firmware Download" : "No",
                        "Lock Key Assigned" : "No",
                        "Failed to get lock key on bootup" : "No",
                        "Lock key has not been backed up" : "No",
                        "Bios was not detected during boot" : "No",
                        "Controller must be rebooted to complete security operation" : "No",
                        "A rollback operation is in progress" : "No",
                        "At least one PFK exists in NVRAM" : "No",
                        "SSC Policy is WB" : "No",
                        "Controller has booted into safe mode" : "No",
                        "Controller shutdown required" : "No"
                },
...
                "BBU_Info" : [
                        {
                                "Model" : "iBBU",
                         ==>    "State" : "Dgd (Needs Attention)",
                                "RetentionTime" : "48 hours +",
                                "Temp" : "29C",
                                "Mode" : "-",
                                "MfgDate" : "2014/02/10",
                                "Next Learn" : "2019/06/27  01:33:42"
                        }
                ]
```

My best guess is that the controller "Needs Attention" because of the battery status, but I can't find documentation for what status=32 means.  Can you point to some info which says that 32 is healthy?

For comparison, here's what MegaCLI says on the same controller:

```
# /opt/MegaRAID/MegaCli/MegaCli64 -AdpBbuCmd -GetBbuStatus -aALL

BBU status for Adapter: 0

BatteryType: iBBU
Voltage: 4014 mV
Current: 0 mA
Temperature: 29 C
[...]

PR #20, opened 2019-12-06
Update storcli.py to work with older cards without ROC temp sensor or…

candlerb:
… Cachevault_info

Also:
* updates URL to storcli documentation and bumps version number
* adds "megaraid_bbu_temperature", analagous to "megaraid_cv_temperature" but for devices with BBU rather than Cachevault
* suppresses "megaraid_battery_backup_healthy" metric for devices where BBU is not present

Fixes #19, #27

Signed-off-by: Brian Candler <b.candler@pobox.com>

PR #32, opened 2019-12-20
ADD storcli detailed CacheVault metrics

ntavares:
There doesn't seem to be any documentation around the "BBU Status" field and, unlike what @mulbc describes on 1b98db9fa72abe93541fb1a7140388504601e303, we experience it now in several states other than [0,32] (16, 64, etc) which we cannot understand; and they last for random multiple days durations.

I therefore followed his suggestion of parsing the detailed output, but only for CacheVault, not BBU. I cannot find now a machine with BBU, so that'll come later.

### DIFF
--- a/storcli.py
+++ b/storcli.py
@@ -147,7 +147,13 @@ def handle_cachevault(args, controller_index, response):
             pass
 
         try:
-            amount, units = search_property(response, 'GasGaugeStatus', 'Capacitance').split(' ', 1)
+            # Another nasty hack as the output format is not consistent for certain values, e.g. 0 and 100
+            capacitance = search_property(response, 'GasGaugeStatus', 'Capacitance')
+            if '%' in capacitance:
+                amount, units = capacitance.split(' ', 1)
+            else:
+                amount = capacitance
+                units = '%'
             add_metric('cv_gas_capacitance', baselabel + ',units="{0}"'.format(units), amount)
         except ValueError:
             pass
--- a/storcli.py
+++ b/storcli.py
@@ -267,8 +267,10 @@ def handle_megaraid_controller(response):
     (controller_index, baselabel) = get_basic_controller_info(response)
 
     # BBU Status value is not explained in the schema, see note in the header.
-    add_metric('bbu_status', baselabel,
-               int(response['Status']['BBU Status']))
+    try:
+        add_metric('bbu_status', baselabel, int(response['Status']['BBU Status']))
+    except:
+        pass
 
     add_metric('degraded', baselabel, int(response['Status']['Controller Status'] == 'Degraded'))
     add_metric('failed', baselabel, int(response['Status']['Controller Status'] == 'Failed'))
--- a/storcli.py
+++ b/storcli.py
@@ -15,6 +15,11 @@ output, i.e.  when you omit the trailing 'J' from the command.
 
 Formatting done with YAPF:
 $ yapf -i --style '{COLUMN_LIMIT: 99}' storcli.py
+
+NOTE about the undocumented "BBU Status":
+As of 201912, there's still an ongoing effort to reverse engineer the value, you can find the latest mappings at:
+https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/issues/27
+(may change to a better location)
 """
 
 from __future__ import print_function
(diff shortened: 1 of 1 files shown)


dswarbrick:
See https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/pull/20

ntavares:
Hi @dswarbrick thanks for sharing! The refered issues are very enlightening in regards to reverse-engineering the BBU Status :-) Indeed I couldn't get feedback from the vendor either but, there seems to be some progress for us!

In any case, this PR adds detailed CV metrics, so we're changing our alerts anyway for the time being. Maybe you want to add the /call/bbu ? :-)

ntavares:
Refactor a bit based on the discussing in #27:
* cv_temperature and battery_backup_healthy now come from the dedicated CV/BBU parser loops, instead of the "/call show all" (handle_megaraid_controller) parser;
* Note that the cvidx parameter is removed: the output on a multiple controller box is NOT a list, which seems to indicate the CV (at least) is indexed to the controller (IOW, one controller can only have one CacheVault);
* I've "introduced" megaraid_bbu_status as the plain value from "BBU Status", with the remark that its interpretation is not yet clear. There's a nice reverse-engineering going on here: https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/issues/27#issuecomment-567884906 , but at least we can start graphing the value and correlate it with other metrics.

@candlerb In the meantime I (finally) found a server with a BBU, so I've added the "detailed" metrics as well. For the BBU, there's lots more of eventually interesting stuff under 'BBU_Firmware_Status', but for now I think it's good.

storcli.py
```diff
@@ -89,17 +266,16 @@ def handle_sas_controller(response):
 def handle_megaraid_controller(response):
     (controller_index, baselabel) = get_basic_controller_info(response)
 
-    # BBU Status Optimal value is 0 for cachevault and 32 for BBU
-    add_metric('battery_backup_healthy', baselabel,
-               int(response['Status']['BBU Status'] in [0, 32]))
+    # BBU Status value is not explained in the schema, see note in the header.
+    add_metric('bbu_status', baselabel,
+               int(response['Status']['BBU Status']))
```
candlerb: Cards with no BBU return the string "NA" for BBU Status, so this needs wrapping with try/except.

storcli.py
```diff
@@ -57,8 +64,178 @@ def main(args):
...
+        return float(number) * 1024
+    return number
+
+def search_property(response_data, section, property):
+    for pair_kv in response_data[section]:
+        if property == pair_kv['Property']:
+            return str(pair_kv['Value']).strip()
+    return ''
+
+# The following is not really TZ aware, deal with caution :-)
```
candlerb: To avoid the TZ problem: take the difference between the learn timestamp and the controller's current clock time, and add this to the system utime.

storcli.py
```diff
@@ -57,8 +64,178 @@ def main(args):
     except KeyError:
         pass
 
+    parsers = {
+        'cv': handle_cachevault,
+        'bbu': handle_bbu
+    }
+
+    for parser, parser_func in parsers.items():
+        data = get_storcli_json('/cALL/' + parser + ' show all J')
```
candlerb: This will be run even if args.no_detailed_bbu is set.  I believe there is sufficient info in the main "all" output for users who don't want detailed info. Examples:

```
                "BBU_Info" : [
                        {
                                "Model" : "iBBU",
                                "State" : "Dgd (Needs Attention)",
                                "RetentionTime" : "48 hours +",
                                "Temp" : "29C",
                                "Mode" : "-",
                                "MfgDate" : "2014/02/10",
                                "Next Learn" : "2019/06/27  01:33:42"
                        }
                ]

```

```
                "Cachevault_Info" : [
                        {
                                "Model" : "CVPM02",
                                "State" : "Optimal",
                                "Temp" : "30C",
                                "Mode" : "-",
                                "MfgDate" : "2014/05/30"
                        }
                ]
```

candlerb:
My general observation about this PR is there seems to be a *lot* of information returned about BBU here and I'm not sure it's all necessary.  It means we probably know more about the BBU than the adapter itself!  To put it another way: does *everything* which appears in the JSON have to be exposed as a metric?

If the state is non-optimal then the user can investigate directly on the card.  e.g. do we really need a separate metric for "microcode update required"?    That state is probably one of the flags in bbu_status anyway, so users who come across this condition will learn what that status value means.

Alternatively: these flags could go as labels into cv_info, in the same way as pd_info has `state="Onln"` or `state="GHS"` as a label, rather than as a separate metric per flag.

OTOH: I agree it is consistent to have cv_info / bbu_info with serial number etc, in the same way as the adapter info.

storcli.py
ntavares: Oh sorry, I missed that one from your commit(s). I'm suspecting this will actually apply to all metrics actually, but we'll see.

storcli.py
ntavares: Yes, that's what I meant in the first bullet of [my other comment](https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/issues/27#issuecomment-568042283), I think it's more clear to have a single branch dealing with the specifics (collaterally covering the "general overview" from "/call show all" that you printed above) than having bbu/cv metrics coming from both codepaths.

The only advantage I see in going back to collecting those 2 metrics is to avoid two more system calls, I just took that as neglictible. In that sense, the option "detailed_bbu" also changed to "no_detailed_bbu". From my experience, the users generally won't mind having more metrics on a new version.

ntavares:
Hi @candlerb, thanks for taking the time to go through this.

> It means we probably know more about the BBU than the adapter itself!

I don't see a problem with that :-) I got curious about the ability to track battery capacitance during its lifetime, it could help predict replacements. 

> That state is probably one of the flags in bbu_status anyway

That might be very true, however until "BBU Status" is completely reverse engineered, no one can tell for sure. Look, I don't even really know what they mean, I just sensed they could point to a problem somehow - and if that's the case, I'd rather have prometheus tell it to me immediatelly rather than having to issue a storcli64 command and dig that out from a huge listing of parameters.

> Alternatively: these flags could go as labels into cv_info, in the same way as pd_info has state="Onln" or state="GHS" as a label, rather than as a separate metric per flag.

Well considering what I know about those metrics, I'd actually have done that, if I had thought about it myself. But we're talking about saving 2 metrics of the whole bunch, and we go back to the need of running storcli64 yourself to figure it out, as I wrote above. I suggest we stick with independent metrics.

dswarbrick:
@ntavares Please rebase on master and resolve any merge conflicts.

dswarbrick:
Closing due to inactivity. Please reopen and resolve merge conflicts if you still wish this to be considered.

PR #34, opened 2020-01-09
Split up yum.sh script long line

SuperQ:
Break up the long pipeline in the yum.sh script to make it easier to
read.

Add bash sanity settings.

Signed-off-by: Ben Kochie <superq@gmail.com>

### DIFF
--- a/yum.sh
+++ b/yum.sh
@@ -6,7 +6,27 @@
 # 
 # Based on apt.sh by Ben Kochie <superq@gmail.com>
 
-upgrades=$(/usr/bin/yum -q check-update | awk 'BEGIN { mute=1 } /Obsoleting Packages/ { mute=0 } mute && /^[[:print:]]+\.[[:print:]]+/ { print $3 }' | sort | uniq -c | awk '{print "yum_upgrades_pending{origin=\""$2"\"} "$1}')
+set -u -o pipefail
+
+filter_awk_script='
+BEGIN { mute=1 }
+/Obsoleting Packages/ {
+  mute=0
+}
+mute && /^[[:print:]]+\.[[:print:]]+/ {
+  print $3
+}
+'
+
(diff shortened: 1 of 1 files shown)


SuperQ:
@candlerb can you validate that this works correctly for you?

yum.sh
```diff
@@ -6,7 +6,26 @@
 # 
 # Based on apt.sh by Ben Kochie <superq@gmail.com>
 
-upgrades=$(/usr/bin/yum -q check-update | awk 'BEGIN { mute=1 } /Obsoleting Packages/ { mute=0 } mute && /^[[:print:]]+\.[[:print:]]+/ { print $3 }' | sort | uniq -c | awk '{print "yum_upgrades_pending{origin=\""$2"\"} "$1}')
+set -u -o pipefail
+
+filter_awk_script="$(cat << 'AWKSCRIPT'
```
candlerb: Subshell just to capture a multi-line literal?  Why not just:

```
filter_awk_script='stuff
goes
here'
```
or even skip the variable:

```
... check-update |
    awk 'stuff
goes
here' |
   sort | ...

File: yum.sh [hunk 2]
```diff
@@ -6,7 +6,26 @@
...
+BEGIN { mute=1 }
+/Obsoleting Packages/ { mute=0 } mute &&
```
candlerb: move 'mute &&' to the next line since it's the two conditions together which are applied:

```
mute && /pattern/ { do stuff }
```

yum.sh
SuperQ: I like to use HERE documents in this way to avoid any escaping issues.

SuperQ:
How about now?

candlerb:
It works: tested on CentOS 6.9, 7.2 and 8.0

PR #35, opened 2020-01-19
apt.sh: return packages pending autoremove

erdnaxe:
I find it useful to have a metric counting APT packages pending autoremove to keep my Debian nodes clean.


dswarbrick:
LGTM.

PR #41, opened 2020-03-04
smartmon.py: fix parse error

paol:
Some device models report "---" in the threshold value where most devices would report "000".
This was crashing the program with a parse exception.


SuperQ (approved):
LGTM, Thanks

SuperQ:
Can you rebase this against the latest version?

PR #45, opened 2020-03-27
nvme_metrics: support for more than 10 devices (Closes #18)

### LINKED ISSUE #18: nvme_metrics breaks on systems with more than 10 NVME drives
https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/blob/0be35b81fccdf74c731d9caacf3ec482af9e972e/nvme_metrics.sh#L53

The line that parses disk name will truncate when it reaches 10 (e.g. /dev/nvme10). This will cause the output to contain duplicate entries with different values.

In my use case, node_exporter will fill my logs with `error gathering metrics... was collected before with the same name and label values` and degrade the scrapes as a result.

herver:
Note: adds a dependency on grep

### DIFF
--- a/nvme_metrics.sh
+++ b/nvme_metrics.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -eu
 
-# Dependencies: nvme-cli, jq (packages)
+# Dependencies: nvme-cli, grep, jq (packages)
 # Based on code from
 # - https://github.com/prometheus/node_exporter/blob/master/text_collector_examples/smartmon.sh
 # - https://github.com/prometheus/node_exporter/blob/master/text_collector_examples/mellanox_hca_temp
@@ -50,7 +50,7 @@ device_list="$(nvme list | awk '/^\/dev/{print $1}')"
 # Loop through the NVMe devices
 for device in ${device_list}; do
   json_check="$(nvme smart-log -o json "${device}")"
-  disk="$(echo "${device}" | cut -c6-10)"
+  disk="$(echo "${device}" | grep -Eo 'nvme[0-9]+')"
 
   # The temperature value in JSON is in Kelvin, we want Celsius
   value_temperature="$(echo "$json_check" | jq '.temperature - 273')"


nvme_metrics.sh
```diff
@@ -50,7 +50,7 @@ device_list="$(nvme list | awk '/^\/dev/{print $1}')"
 # Loop through the NVMe devices
 for device in ${device_list}; do
   json_check="$(nvme smart-log -o json "${device}")"
-  disk="$(echo "${device}" | cut -c6-10)"
+  disk="$(echo "${device}" | grep -Eo 'nvme[0-9]+')"
```
dswarbrick, line 53: You could also just specify the field delimiter to `cut`
```suggestion
  disk="$(echo "${device}" | cut -d/ -f3)"
```
Proof:
```
$ echo /dev/nvme0n1 | cut -d/ -f3
nvme0n1
```
Otherwise, if we're not concerned about "wild" dependencies, the _right_ way to do this is with `basename`:
```
$ basename /dev/nvme0n1
nvme0n1
```
Since `basename` is part of the GNU coreutils suite of tools, I would be _highly_ surprised if it is not present on a system.

RichiH:
GNU would not be available on OS X or BSD. I think basename is beyond just
GNU though.

I agree that basename and -f3 are The Unix Way.

Sent by mobile; please excuse my brevity.

dswarbrick:
`basename` is part of POSIX.

basename on FreeBSD: https://www.freebsd.org/cgi/man.cgi?query=basename&sektion=1&manpath=FreeBSD+12.1-RELEASE+and+Ports

basename on OpenBSD: https://man.openbsd.org/basename.1

I would expect other BSDs such as NetBSD to be the same. Apple seems to have taken down macOS HTML man pages, but cached versions seem to suggest that it has `basename` (again, as expected).

bdrung:
There is a Shell built-in way to calculate the basename:
```bash
${device##*/}
```

bdrung:
Proposed this solution in pull request #76.

dswarbrick:
Obsoleted by #76.

PR #50, opened 2020-04-11
add hddtemp collector

andresbott:
this textfile collector uses hddtempd to get the disks and disk temperatures, and expose as prometheus metric with one label per disk.

### DIFF
--- a/hddtemp.sh
+++ b/hddtemp.sh
@@ -4,6 +4,11 @@
 # Description: Expose hard dist temperature metrics using hddtempd
 # this expects a running and configured hddtempd,
 # refer to t he hddtemp documentation for this case.
+# 
+# if disks are repored as in sleep ("SLP") the value will be -101
+# for all other unknown status, the reported value will be -100
+# this scenario asumes that any disk will never run in -100 degree celcius (-148°F)
+#
 #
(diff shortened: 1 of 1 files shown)


dswarbrick:
Isn't this somewhat redundant, given that there are already two SMART collectors in this repo, which are capable of fetching HDD temperatures (and much more)?

AndresBott:
maybe ... i haven't seen the smart collector before.
I will try it out to see if it is a better fit.

AndresBott:
After a quick check on my server smartmon.sh seems to completely cover the temperature feature.
The only drawbacks for smartmon.sh is that the the script needs to run as root, as where as hddtempd can be read as a normal user.

I'll let you decide if you want to merge or not, I guess since it is already covered by smart we can close this.

dswarbrick:
Whilst this script might not require root privileges, the hddtempd daemon that it talks to _does_ need to run as root (and I expect that at some level, is sending the same SMART commands via SCSI generic ioctl that `smartctl` does). A sudo wrapper script would also arguably not require root access.

Given the number of unfixed typos in this PR, and the fact that the smartmon collector already fulfills this niche... closing.

PR #55, opened 2020-05-29
smartmon.py: Fixed check if device is active.

lomion0815:
smartctl --nockeck standby throughs an exception with errorcode "2" in case the device is not active. Since this exception or the errorcode is not passed to the calling function the device was recognized to be active in this case. Fixed it by re-raising exception.


dswarbrick:
@lomion0815 Please verify whether this was already resolved by the aforementioned PR, and fix the merge conflicts if you still wish to submit this.

dswarbrick:
Closing due to inactivity and unresolved merge conflicts.

PR #58, opened 2020-06-29
have yum.sh deal with long lines from yum

britcey:
When yum output is redirected, it wraps to 80 columns (https://bugzilla.redhat.com/show_bug.cgi?id=584525) - for long lines, this results in yum_upgrades_pending metrics with origin="".

Adding xargs to the pipeline corrects things so awk works as intended.

Signed-off-by: Ben Ritcey <ben+github@ritcey.com>


britcey:
@SuperQ does this look like a sane change?

PR #63, opened 2020-09-02
pacman: Fix script failure for no updates

veecue:
When no updates are available, checkupdates fails with exit code 2. Catch this by wrapping the command using || true.


pacman.sh
```diff
@@ -14,10 +14,10 @@ set -o pipefail
 
 if [ -x /usr/bin/checkupdates ]
 then
-    updates=$(/usr/bin/checkupdates | wc -l)
+    updates=$( (checkupdates || true) | wc -l)
```
hansmi, line 17: `|| true` ignores all failures. As-is it would even pick another `checkupdates` executable if found in `$PATH`. Isn't there a specific exit code? If need be the output can also be checked.

PR #67, opened 2020-10-15
smartmon.py: Escape quotes in label values

hansmi:
Disk names can contain quotes, e.g. in "Toshiba 2.5" HDD MK..52GSX".


dswarbrick:
LGTM. Note that @SuperQ and I are considering requiring Python collectors to use the official Prometheus https://github.com/prometheus/client_python module, which will avoid / abstract this kind of thing in future.

PR #71, opened 2020-10-23
Update ipmitool

troymuller:
adding in support for 'percent' as some Fans and other devices (depending on hardware) appear as percentage instead of as a factor of speed.


troymuller:
adding a signed one instead

PR #72, opened 2020-10-23
Adding percent to exporter

troymuller:
adding in support for 'percent' as some Fans and other devices (depending on hardware) appear as percentage instead of as a factor of speed.


troymuller:
master, which didn't have the change

PR #74, opened 2020-10-26
mellanox_hca_temp: Rename hca_device label to device

bdrung:
To be consistent with the `node_infiniband_*` labels (e.g. `node_infiniband_info`) from the Prometheus node exporter, rename the
metric label `hca_device` to `device`. Then you can write an alert to check for missing temperature metrics:

```
node_infiniband_info unless on (instance, device) node_infiniband_hca_temp_celsius
```

Signed-off-by: Benjamin Drung <benjamin.drung@cloud.ionos.com>


dswarbrick:
LGTM

PR #75, opened 2020-11-09
smartmon.py: add disk property for multi-disk devices

### LINKED ISSUE #23: Issues with smartmon.py
1. smartmon.py does not return the "raw" values of metrics (only cooked value/threshold/worst): but almost always the raw values are the meaningful ones.  e.g.

```
# HELP smartmon_attr_value SMART metric attr_value
# TYPE smartmon_attr_value gauge
smartmon_attr_value{name="reallocated_sector_ct",disk="/dev/sda"} 100
smartmon_attr_value{name="power_on_hours",disk="/dev/sda"} 97
smartmon_attr_value{name="power_cycle_count",disk="/dev/sda"} 99
smartmon_attr_value{name="wear_leveling_count",disk="/dev/sda"} 87
smartmon_attr_value{name="erase_fail_count_total",disk="/dev/sda"} 100
smartmon_attr_value{name="airflow_temperature_cel",disk="/dev/sda"} 70
smartmon_attr_value{name="total_lbas_written",disk="/dev/sda"} 99
```

None of those are meaningful.  For example, the power-on time is not 97 hours, and the temperature is not 70 degrees.  You have to trust that the nearer to zero it is, the worse it is.

Compare smartmon.sh:

```
smartmon_airflow_temperature_cel_raw_value{disk="/dev/sda",type="sat",smart_id="190"} 3.000000e+01
smartmon_airflow_temperature_cel_value{disk="/dev/sda",type="sat",smart_id="190"} 70
smartmon_erase_fail_count_total_raw_value{disk="/dev/sda",type="sat",smart_id="182"} 0.000000e+00
smartmon_erase_fail_count_total_value{disk="/dev/sda",type="sat",smart_id="182"} 100
smartmon_power_cycle_count_raw_value{disk="/dev/sda",type="sat",smart_id="12"} 6.000000e+00
smartmon_power_cycle_count_value{disk="/dev/sda",type="sat",smart_id="12"} 99
smartmon_power_on_hours_raw_value{disk="/dev/sda",type="sat",smart_id="9"} 1.094000e+04
smartmon_power_on_hours_value{disk="/dev/sda",type="sat",smart_id="9"} 97
smartmon_reallocated_sector_ct_raw_value{disk="/dev/sda",type="sat",smart_id="5"} 0.000000e+00
smartmon_reallocated_sector_ct_value{disk="/dev/sda",type="sat",smart_id="5"} 100
smartmon_total_lbas_written_raw_value{disk="/dev/sda",type="sat",smart_id="241"} 4.342395e+11
smartmon_total_lbas_written_value{disk="/dev/sda",type="sat",smart_id="241"} 99
smartmon_wear_leveling_count_raw_value{disk="/dev/sda",type="sat",smart_id="177"} 2.670000e+02
smartmon_wear_leveling_count_value{disk="/dev/sda",type="sat",smart_id="177"} 87
```

The cooked values are there too, but the raw ones are the useful ones (e.g. airflow temperature 30 degrees, power on hours 10940)

Aside: using named metrics like this is I think also better practice than having a shared metric "smartmon_attr_value" containing unrelated metrics (although since they are all cooked it's arguably OK)

2. With drives downstream of a megaraid card, smartmon.py doesn't give unique labels to each metric:

```
# HELP smartmon_device_active SMART metric device_active
# TYPE smartmon_device_active gauge
smartmon_device_active{disk="/dev/bus/0"} 1
smartmon_device_active{disk="/dev/bus/0"} 1
smartmon_device_active{disk="/dev/bus/0"} 1
smartmon_device_active{disk="/dev/bus/0"} 1
smartmon_device_active{disk="/dev/bus/0"} 1
...
```

Compare with smartmon.sh:

```
# HELP smartmon_device_active SMART metric device_active
# TYPE smartmon_device_active gauge
smartmon_device_active{disk="/dev/bus/0",type="sat+megaraid,10"} 1
smartmon_device_active{disk="/dev/bus/0",type="sat+megaraid,11"} 1
smartmon_device_active{disk="/dev/bus/0",type="sat+megaraid,12"} 1
smartmon_device_active{disk="/dev/bus/0",type="sat+megaraid,13"} 1
smartmon_device_active{disk="/dev/bus/0",type="sat+megaraid,14"} 1
...
```

-----

WORKAROUND:  use smartmon.sh instead.  (Bonus: it also works on older systems that don't have python3)

lheckemann:
Fixes #23 I think

### DIFF
--- a/smartmon.py
+++ b/smartmon.py
@@ -86,8 +86,8 @@ class Device(collections.namedtuple('DeviceBase', 'path opts')):
     @property
     def base_labels(self):
         if '+' in self.type:
-            return {'dev': self.path, 'disk': self.type.partition("+")[2]}
-        return {'disk': self.path}
+            return {'device': self.path, 'disk': self.type.partition("+")[2]}
+        return {'device': self.path}
 
     def smartctl_select(self):
         return ['--device', self.type, self.path]
--- a/smartmon.py
+++ b/smartmon.py
@@ -85,7 +85,7 @@ class Device(collections.namedtuple('DeviceBase', 'path opts')):
 
     @property
     def base_labels(self):
-        return {'device': self.path, 'disk': self.type.partition("+")[2] or "0"}
+        return {'device': self.path, 'disk': self.type.partition('+')[2] or '0'}
 
     def smartctl_select(self):
         return ['--device', self.type, self.path]
--- a/smartmon.py
+++ b/smartmon.py
@@ -85,6 +85,8 @@ class Device(collections.namedtuple('DeviceBase', 'path opts')):
 
     @property
     def base_labels(self):
+        if '+' in self.type:
+            return {'dev': self.path, 'disk': self.type.partition("+")[2]}
         return {'disk': self.path}
 
     def smartctl_select(self):
--- a/smartmon.py
+++ b/smartmon.py
@@ -85,9 +85,7 @@ class Device(collections.namedtuple('DeviceBase', 'path opts')):
 
     @property
     def base_labels(self):
-        if '+' in self.type:
-            return {'device': self.path, 'disk': self.type.partition("+")[2]}
-        return {'device': self.path}
+        return {'device': self.path, 'disk': self.type.partition("+")[2] or "0"}
 
     def smartctl_select(self):
         return ['--device', self.type, self.path]


SuperQ:
Is this going to cause inconsistent labels? We want to make sure that we return the same label set on the same metrics.

Perhaps we should always return `{device="/dev/foo",disk="/dev/foo"}` for single-disk devices.

lheckemann:
I'm not a fan of specifically duplicating the disk label, that's likely to be confusing. My actual preference would be changing it to `device` being the block device path (/dev/sda) and `disk` being the disk index; but that will break existing use. Another alternative would be concatenating the disk index with the block device path, forming a name like `/dev/sda+12`. Not really happy with any of the backwards-compatible solutions I can think of though.

SuperQ:
What if did a breaking change to make the labels match the node_exporter's node_disk collector? It uses `{device="sda"}`. That would make the most sense to me.

Breaking changes are ok, as long as we have a good reason. Matching the node_exporter labels is a good reason.

smartmon.py
```diff
@@ -85,7 +85,9 @@ def type(self):
 
     @property
     def base_labels(self):
-        return {'disk': self.path}
+        if '+' in self.type:
+            return {'device': self.path, 'disk': self.type.partition("+")[2]}
+        return {'device': self.path}
```
SuperQ, line 90: This also need a disk label, something static like `'disk': '0'` is fine.

smartmon.py
```diff
@@ -85,7 +85,7 @@ def type(self):
 
     @property
     def base_labels(self):
-        return {'disk': self.path}
+        return {'device': self.path, 'disk': self.type.partition("+")[2] or "0"}
```
dswarbrick, line 88: nit: standardise on single quotes for keywords, machine-readable strings etc, and only use double quotes for natural language strings (since they may contain apostrophes).

SuperQ (approved):
LGTM, Thanks!

PR #76, opened 2020-11-19
nvme_metrics.sh: Support for more than 10 NVMe and use JSON output of nvme

### LINKED ISSUE #18: nvme_metrics breaks on systems with more than 10 NVME drives

bdrung:
The `cut` command assumes that `device` is in the format `/dev/nvmeX` and will fail if the NVMe number is higher than 9 (e.g. `/dev/nvme10`).

Therefore use a built-in shell method to calculate the basename of the device. (Closes #18)

Parsing the human-readable output of nvme with awk is fragile since the output format is not promised to be stable. Therefore use the JSON output format to make the script more robust.

### DIFF
--- a/nvme_metrics.sh
+++ b/nvme_metrics.sh
@@ -50,7 +50,7 @@ device_list="$(nvme list | awk '/^\/dev/{print $1}')"
 # Loop through the NVMe devices
 for device in ${device_list}; do
   json_check="$(nvme smart-log -o json "${device}")"
-  disk="$(echo "${device}" | cut -c6-10)"
+  disk="${device##*/}"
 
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
Nice, thanks.

PR #79, opened 2020-12-28
Add flake8 linting for Python

SuperQ:
Add flake8 CircleCI to lint Python scripts.
* Add TODO tw_cli.py, not Python3 compatible.

Signed-off-by: Ben Kochie <superq@gmail.com>

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,9 @@
+[flake8]
+import-order-style = google
+max-line-length = 100
+
+exclude=
+  .git,
+  .circleci,
+  # TODO: tw_cli.py needs to be updated for Python 3.
+  tw_cli.py
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,9 +8,21 @@ executors:
   shellcheck:
     docker:
     - image: koalaman/shellcheck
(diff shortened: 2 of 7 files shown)


SuperQ:
CC @dswarbrick

.flake8
```diff
@@ -0,0 +1,8 @@
+[flake8]
```
brian-brazil, line 1: We should probably be consistent for this line: https://github.com/prometheus/client_python/blob/master/tox.ini#L83

SuperQ: Makes sense, done.

dswarbrick:
Nice, I am hugely in favour of this. We do some pretty strict Python linting where I work, including https://github.com/psf/black, which kinda brings Python code into the world of `go fmt`.

btw, I can prepare a PR for tw_cli.py if desired - we apply a debian patch to this in the debian package, to make it python3 compatible.

SuperQ:
Yes, it would be appreciated if you could update tw_cli.py.

dswarbrick:
> Yes, it would be appreciated if you could update tw_cli.py.

See #80

PR #80, opened 2020-12-29
Update tw_cli.py to Python 3

dswarbrick:
Convert `tw_cli.py` textfile collector to Python 3 and address flake8 nags. Also do a few minor cleanups to docstrings, formatting, and simplify some if-statements.

Note that this script still uses `OptionParser`, which has been deprecated since Python 3.2. We could convert this to `ArgumentParser`, but this would raise our minimum supported Python version to 3.2 (which is still ancient). @SuperQ let me know if you want me to go ahead with that. The smartmon.py and storcli.py collectors already use `ArguementParser`, so one could "argue" that we already require Python 3.2 ;-)

This script still contains a few references to Nagios and associated terminology, as it was originally based on a very old Nagios plugin. Overall it could be substantially simplified and made more elegant by using modern Python 3 techniques, but without a 3ware card to test with, or a reliable way to mock the output of `tw_cli`, I am hesitant to do so.

According to Broadcom documentation, `storcli` can also be used to manage 3ware cards, so I'm not sure how much effort it is worth putting into this collector, considering that we have the storcli collector in reasonably good shape (and the JSON output of `storcli` is much more robust to parse).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,5 +5,3 @@ max-line-length = 100
 exclude=
   .git,
   .circleci,
-  # TODO: tw_cli.py needs to be updated for Python 3.
-  tw_cli.py
--- a/tw_cli.py
+++ b/tw_cli.py
@@ -31,14 +31,14 @@ METRIC_PREFIX = 'tw_cli'
 
 
 def exit_error(msg):
-    print(METRIC_PREFIX + '_cli_error{message="' + str(msg) + '"}\t1')
+    print('{}_cli_error{{message="{}"}}\t1'.format(METRIC_PREFIX, msg))
     sys.exit(1)
 
 
 def exit_clean():
     global METRICS
     for mk, mv in METRICS.items():
-        print(METRIC_PREFIX + '_' + mk + '\t' + str(mv))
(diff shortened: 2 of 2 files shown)


SuperQ:
I'm fine with anything >= Python 3.4

SuperQ:
On the topic of updating these scripts, I was thinking about having the Python scripts use the official Prometheus client_python for rendering.

The Promtheus Python has a write_to_textfile method, we could do something similar with `print(generate_latest(registry))`.

dswarbrick:
`OptionParser` converted to `ArgumentParser`. This should now be ready to merge.

@SuperQ I think that using the official client_python is a good idea (in the medium to long term), as it will eliminate a lot of manual exposition format rendering, which varies greatly in quality from one collector to the next.

SuperQ (approved):
LGTM, Thanks!

SuperQ:
Should we squash the commits here, or merge as-is?

dswarbrick:
I'll leave that up to you. I merely pushed them individually to make it easier for you to review. The changes are much more extensive than the simple python 2 to python 3 patch that we apply in the debian package. Whilst fixing the flake8 nags, I was also cleaning / simplifying a lot of other code.

PR #88, opened 2021-01-23
btrfs_stats: detect empty lines in Python 3

dkess:
The `line == ''` check doesn't work in Python 3, which uses bytestrings. Replace it with `not line`.

Without this change, the script can crash, since the empty line will trigger the RuntimeError on line 49.


dswarbrick:
Nice catch! Please sign off your commit and I'll be happy to merge this.

PR #89, opened 2021-02-07
lvm-prom-collector: Fix text format error

sr:
This fixes the following error:

```
level=error ts=2021-02-07T10:44:19.990Z caller=textfile.go:209 collector=textfile msg="failed to collect textfile data" file=metrics.prom err="failed to parse textfile data from \"/var/lib/node-exporter/lvm.prom\": text format parsing error in line 15: expected float as value, got \"\""
```

For deactivated thin LVs, the `data_percent` field is empty, which causes the script to emit a metric without a value.

This handles this case by simply not emitting `node_lvm_thin_pools_allocated` if `data_percent` is empty.

Another approach could be to filter out inactive LVs, like this:

```
lvs --noheadings --select 'lv_attr=~[^t.*] && lv_active=active' --units b --nosuffix --unquoted --nameprefixes --options lv_uuid,vg_name,lv_name,data_percent
```

However this works just fine and is probably easier to understand.

/cc @badrabubker


dswarbrick:
@sr Can you please rebase this on current master, and confirm whether it is still an issue?

PR #98, opened 2021-06-08
smartmon.py: look for smartctl just in case it is not in path

bufanda:
When used in a cronjob the path environment variable might not be set, so the script will look in the most common pathes for the binary. If it is not found it will be used as is, so script might fail in this case.

### DIFF
--- a/smartmon.py
+++ b/smartmon.py
@@ -9,6 +9,8 @@ import shlex
 import subprocess
 import sys
 
+from os import listdir, path
+
 device_info_re = re.compile(r'^(?P<k>[^:]+?)(?:(?:\sis|):)\s*(?P<v>.*)$')
 
 ata_error_count_re = re.compile(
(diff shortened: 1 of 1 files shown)


dswarbrick:
Wouldn't it be a lot simpler to just fix the PATH env var in the cronjob?

bufanda:
> Wouldn't it be a lot simpler to just fix the PATH env var in the cronjob?

Every cron has only a limited PATH variable and sure I could set the PATH variable for the cronjob, but then imo it should be at least be documented somewhere, because debugging cronjob is not an easy task, and this would simply avoid that hassle.

dswarbrick:
The proposed functionality essentially duplicates what normal shell PATH behaviour does, albeit hard-coding a list of paths in the script. This sort of thing does not belong in textfile collectors.

Fix your PATH environment, or alternatively I would accept a PR which added a CLI argument for specifying the exact path to the `smartctl` binary.

PR #100, opened 2021-08-29
smartmon.sh: add label rotational

yackushevas:
For some alerts, it is important to understand what type of disk - hdd or ssd.


dswarbrick:
Is this still needed? node_exporter provides an equivalent metric natively, e.g:
```
# HELP node_disk_ata_rotation_rate_rpm ATA disk rotation rate in RPMs (0 for SSDs).
# TYPE node_disk_ata_rotation_rate_rpm gauge
node_disk_ata_rotation_rate_rpm{device="sda"} 7200
```

PR #102, opened 2021-09-19
add xentop text collector script

oe1rfc:
This adds a script to parse the values of [xentop](https://wiki.xenproject.org/wiki/Xentop(1)) as prometheus metrics.

gawk is probably needed for `FPAT`, which is used because xentop reports unset limits as `no limit`, which breaks whitespace-based field separation.

### DIFF
--- a/xentop
+++ b/xentop
@@ -1,0 +1,60 @@
+#!/usr/bin/awk -f
+# xentop -i 1 -f -b | ./xentop
+
+function export_metric(metric) {
+	printf("\n# HELP %s%s %s from xentop\n", namespace, metric, metric_fields[metric]);
+	printf("# TYPE %s%s gauge\n", namespace, metric);
+	for (i=0; i < dom_count; i++) {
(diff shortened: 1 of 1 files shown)


dswarbrick:
Are you aware of https://github.com/bwesterb/go-xentop, which includes a sample Prometheus exporter?

oe1rfc:
@dswarbrick I think at the time of writing this I was not aware of [go-xentop](https://github.com/bwesterb/go-xentop), but I'm also not sure if it is still a maintained project since it depends on a `prometheus/client_golang` client release from 2019.
Personally I think that having to run a webserver as root on a virtualization host for statistics is not a good idea, but I can close this PR if there is no interest in it.

dswarbrick:
It would probably be pretty trivial to enhance go-xentop to optionally use sudo to call xentop, to avoid running the exporter as root.

Is there no way of getting machine-readable stats from Xen, e.g. in JSON format? Parsing the human-readable output of utils is often pretty fragile, as cosmetic changes to the output format can often break such approaches.

PR #106, opened 2021-11-21
Create systemd_failed_units

r-darwish:


r-darwish:
Didn't notice that the exporter has a systemd collector

PR #108, opened 2021-12-16
Add node_os_info.sh collector

candlerb:
Enable older OSes, like CentOS 6 and early versions of Ubuntu 12.04, to generate `node_os_info` and `node_os_version` metrics.

See https://github.com/prometheus/node_exporter/issues/2252

### DIFF
--- a/node_os_info.sh
+++ b/node_os_info.sh
@@ -1,0 +1,58 @@
+#!/bin/sh -e
+
+# Generate node_os_info and node_os_version metrics on legacy systems
+# which are not handled by node_exporter's own collector
+# (e.g. CentOS 6)
+
+[ -f /etc/os-release ] && exit 0
+[ -f /usr/lib/os-release ] && exit 0
+
+ID=""
+ID_LIKE=""
+NAME=""
+PRETTY_NAME=""
+VERSION=""
+VERSION_CODENAME=""
+VERSION_ID=""
+VERSION_METRIC=""
+
+if [ -f /etc/redhat-release ]; then
+  # CentOS release 6.10 (Final)
+  PRETTY_NAME="$(cat /etc/redhat-release)"
+  if [ -f /etc/centos-release ]; then
+    ID="centos"
+  elif [ -f /etc/oracle-release ]; then
+    ID="ol"
+  fi
+  ID_LIKE="rhel fedora"
+  NAME="$(expr "$PRETTY_NAME" : '\([^ ]*\)')" || true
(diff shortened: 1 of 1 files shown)


node_os_info.sh
```diff
@@ -0,0 +1,58 @@
...
+  VERSION_ID="$(expr "$PRETTY_NAME" : '.* \([0-9][0-9.]*\)')" || true
+  # metric cannot distinguish 6.1 from 6.10, so only keep the integer part
+  VERSION_METRIC="$(expr "$VERSION_ID" : '\([0-9]*\)')" || true
+elif [ -f /etc/lsb-release ]; then
+  # DISTRIB_ID=Ubuntu
+  # DISTRIB_RELEASE=12.04
+  # DISTRIB_CODENAME=precise
+  # DISTRIB_DESCRIPTION="Ubuntu 12.04 LTS"
+  # Beware, old versions of CentOS with package "redhat-lsb-core" look like this instead:
+  # LSB_VERSION=base-4.0-amd64:base-4.0-noarch:core-4.0-amd64:core-4.0-noarch
```
SuperQ, line 39: Should this be in the section above?

candlerb: I put it here because this is the section which parses `/etc/lsb-release`.

This is what I saw on a couple of machines:

```
[root@r5 ~]# cat /etc/lsb-release
LSB_VERSION=base-4.0-amd64:base-4.0-noarch:core-4.0-amd64:core-4.0-noarch
[root@r5 ~]# ls -l /etc/lsb-release.d/
total 0
-rw-r--r--. 1 root root 0 Feb 22  2013 base-4.0-amd64
-rw-r--r--. 1 root root 0 Feb 22  2013 base-4.0-noarch
-rw-r--r--. 1 root root 0 Feb 22  2013 core-4.0-amd64
-rw-r--r--. 1 root root 0 Feb 22  2013 core-4.0-noarch
[root@r5 ~]# cat /etc/redhat-release
CentOS release 6.4 (Final)
```

RedHat's `lsb-release` technically implements the [spec](https://linux.die.net/man/1/lsb_release), but is completely useless because it misses out all the "optional" fields which contain the useful info.

This is why I now test for `/etc/redhat-release` before `/etc/lsb-release` (originally I had them the other way round)

SuperQ: Makes sense. Thanks.

SuperQ (approved):
Most of our scripts are bash, but I'm also OK with basic POSIX sh.

PR #113, opened 2022-02-11
Add NVME support to smartmon.py

### LINKED ISSUE #110: smartmon.py won't collect nvme drive metrics
it only collect sata sd* drives metrics

psi-4ward:
TBDC: NVME-Metrics are more like `nvme_metrics.sh` like than `smartmon.py`. 

closes #110

### DIFF
--- a/smartmon.py
+++ b/smartmon.py
@@ -355,7 +355,7 @@ def collect_nvme_metrics(device):
         elif label == 'Power On Hours':
             yield Metric('power_on_hours_total', device.base_labels, value.replace(',', ''))
         elif label == 'Temperature':
-            yield Metric('temperature_celcius', device.base_labels, value.replace(' Celsius', ''))
+            yield Metric('temperature_celsius', device.base_labels, value.replace(' Celsius', ''))
         elif label == 'Unsafe Shutdowns':
             yield Metric('unsafe_shutdowns_total', device.base_labels, value)
         elif label == 'Media and Data Integrity Errors':
--- a/smartmon.py
+++ b/smartmon.py
@@ -203,6 +203,11 @@ def device_smart_capabilities(device):
             (bool): True whenever SMART is available, False otherwise.
             (bool): True whenever SMART is enabled, False otherwise.
     """
+
+    # NVME devices are SMART capable
+    if device.type == 'nvme':
+        return True, True
+
     groups = device_info(device)
 
     state = {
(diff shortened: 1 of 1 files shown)


smartmon.py
```diff
@@ -325,6 +330,44 @@ def collect_ata_error_count(device):
...
+        elif label == 'Available Spare Threshold':
+            yield Metric('available_spare_threshold_ratio', device.base_labels, value[0:-1])
+        elif label == 'Percentage Used':
+            yield Metric('percentage_used_ratio', device.base_labels, value[0:-1])
+        elif label == 'Power Cycle':
+            yield Metric('power_cycles_total', device.base_labels, value)
+        elif label == 'Power On Hours':
+            yield Metric('power_on_hours_total', device.base_labels, value.replace(',', ''))
+        elif label == 'Temperature':
+            yield Metric('temperature_celcius', device.base_labels, value.replace(' Celsius', ''))
```
lheckemann, line 358:
```suggestion
            yield Metric('temperature_celsius', device.base_labels, value.replace(' Celsius', ''))
```

dswarbrick:
The smartmon.py script has recently undergone non-trivial refactoring to use the official Prometheus client_python library, and this PR would need to be updated accordingly.

I still feel that the best place to implement NVMe metrics collection is in the nvme_metrics collector. Yes, `smartctl` now supports NVMe also. However, `nvme-cli` is the de facto tool for managing / querying NVMe devices, and goes a lot further than `smartctl`, since it also supports NVMeoF devices and vendor extensions which likely won't land in `smartctl` anytime soon (or ever).

If we implement NVMe metrics in two completely different collectors, this is going to be confusing for users. I expect that many setups will run both collectors anyway, which would essentially result in duplicated metrics if the `smartmon` collector also gathered NVMe metrics. I think it's best to leave `smartctl` to do what it is best at - handling ATA / SCSI / SAS drives and hardware RAID controllers.

The only way I would condone adding NVMe metrics to the `smartmon` collector would be if it were behind a CLI switch, and disabled by default.

@SuperQ WDYT?

SuperQ:
> If we implement NVMe metrics in two completely different collectors, this is going to be confusing for users.

We can provide more than one way to do things in this repo.

> The only way I would condone adding NVMe metrics to the smartmon collector would be if it were behind a CLI switch, and disabled by default.

This is probably a good idea.

dswarbrick:
Closing due to inactivity and unresolved merge conflicts.

PR #121, opened 2022-06-25
fixed issue when run from crontab

saeed-mcu:
Signed-off-by: Saeed <sam137115@gmail.com>

when run from crontab lvs , pvs , vgs command make error and exit code 127 which means command not found
and if i use full path for command everything works well

### DIFF
--- a/lvm-prom-collector
+++ b/lvm-prom-collector
@@ -79,7 +79,7 @@ if [ "$physical" = true ] ; then
   echo "# HELP node_physical_volume_free Physical volume free space in bytes"
   echo "# TYPE node_physical_volume_free gauge"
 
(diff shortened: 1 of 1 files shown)


dswarbrick:
If these commands were failing as "not found", this merely suggests that your cron environment is lacking `/usr/sbin` in its path, which I find a bit odd.

You could merely add an explicit `PATH` in your cron entry which calls this textfile collector, but again, I find it quite strange that this would be needed.

dswarbrick:
Closing due to inactivity. Also, hardcoded paths in scripts generally cause more problems than they resolve. Fix you PATH in your crontab if this is still an issue for you.

PR #99, opened 2021-07-17
Add new apt_info.py script

kyrofa:
Currently, apt information is extracted via a shell script by way of `apt-get --just-print dist-upgrade`, and it has a number of issues. For example, it includes held packages in the metrics, and includes information about packages that aren't even installed. This PR fixes #85 and fixes #90 by replacing `apt.sh` with a new Python-based script that uses the APT API to extract everything we need, while fixing known issues and remaining mostly compatible with the old metric format.

PR #122, opened 2022-07-12
Added Security Updates as separate Value

manuel-held:
Signed-off-by: Manuel <manuel@held.is>

### DIFF
--- a/apt.sh
+++ b/apt.sh
@@ -19,6 +19,8 @@ autoremove="$(/usr/bin/apt-get --just-print autoremove \
   | /usr/bin/awk '/^Remv/{a++}END{printf "apt_autoremove_pending %d", a}'
 )"
 
+security_upgrades="$(apt list --upgradable 2>&1 | grep "\-security" | wc -l)"
+
 echo '# HELP apt_upgrades_pending Apt package pending updates by origin.'
 echo '# TYPE apt_upgrades_pending gauge'
 if [[ -n "${upgrades}" ]] ; then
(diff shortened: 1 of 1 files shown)


dswarbrick:
Isn't this already achievable with simple label matching? For example, with these source metrics:
```
# HELP apt_upgrades_pending Apt package pending updates by origin.
# TYPE apt_upgrades_pending gauge
apt_upgrades_pending{origin="Debian:11.4/stable",arch="all"} 2
apt_upgrades_pending{origin="Debian:11.4/stable",arch="amd64"} 14
apt_upgrades_pending{origin="Debian:11.4/stable,Debian-Security:11/stable-security",arch="all"} 2
apt_upgrades_pending{origin="Debian:11.4/stable,Debian-Security:11/stable-security",arch="amd64"} 11
apt_upgrades_pending{origin="Debian-Security:11/stable-security",arch="amd64"} 1
```
... I can get the number of security upgrades with `sum(apt_upgrades_pending{origin=~".*Debian-Security.*"})`

PS: I think we'd prefer not to tempt fate by doing precisely the thing that `apt` warns against doing.
```
$ apt list --upgradable | grep -c "\-security"

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

manuel-held:
Hi @dswarbrick, yes this is possible. But if you're using Grafana and use a Query like `sum by(job) (apt_upgrades_pending{origin=~".*Security.*"})` you get "No Data" if there are no Updates, which is not so informative.

dswarbrick:
@manuel-held How about simply changing your query to
```
sum by(job) (apt_upgrades_pending{origin=~".*Security.*"}) or 0
```

SuperQ:
This has been superseded by https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/pull/99

copolycube:
Hello @dswarbrick , 
with `sum by(job) (apt_upgrades_pending{origin=~".*Security.*"}) or vector(0)` you only get `{} 0` as a global result when there are no match / no security upgrades on each instance.
Is it possible to still have a vector with `0` for each original value of `apt_upgrades_pending` instead ?

PR #126, opened 2022-08-06
Fix copy-pasted smart ids

GunArm:
Signed-off-by: GunArm <james.barrett.lewis@gmail.com>

### DIFF
--- a/smartmon.sh
+++ b/smartmon.sh
@@ -98,9 +98,9 @@ parse_smartctl_scsi_attributes() {
   [ -n "$power_on" ] && echo "power_on_hours_raw_value{${labels},smart_id=\"9\"} ${power_on}"
   [ -n "$temp_cel" ] && echo "temperature_celsius_raw_value{${labels},smart_id=\"194\"} ${temp_cel}"
   [ -n "$lbas_read" ] && echo "total_lbas_read_raw_value{${labels},smart_id=\"242\"} ${lbas_read}"
-  [ -n "$lbas_written" ] && echo "total_lbas_written_raw_value{${labels},smart_id=\"242\"} ${lbas_written}"
+  [ -n "$lbas_written" ] && echo "total_lbas_written_raw_value{${labels},smart_id=\"241\"} ${lbas_written}"
   [ -n "$power_cycle" ] && echo "power_cycle_count_raw_value{${labels},smart_id=\"12\"} ${power_cycle}"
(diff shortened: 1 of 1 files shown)


dswarbrick:
@GunArm Thanks for submitting this. The fixed total LBAs written ID aligns with my research. Do you have a reference for the grown defects list ID? I was not aware that SMART IDs could be negative.

GunArm:
Sorry, surprised I didn't post any accompanying comments.

SCSI drives don't have SMART metrics at all, but the bent of this script is smart metrics.  So (before my PR) there is some fudging in this script for SCSI drives to get what data from them is possible, and then put it into the fields where it would have gone if they were ATA drives with SMART.  But for example there is no "smart id #194" on a scsi drive, there is just the temperature read in a different way, and the script (I presume for consistency of storage between drive types) formats it as if it was an ATA with smart metrics.

ATA drives with SMART will output Reallocated Sectors and Pending Sectors which are both related to bad sector detection.  Sata drives make available only "Grown defects", which (as far as I can tell) is the sum of the above two (or really the above two are a more granular breakout of the defect count).  Point being there is no SMART metric for "grown defects" since it's what scsi drives (without smart) call the defect count.  If it were possible to split up the scsi's "grown defects" to populate the existing two SMART fields for defects, that would be better, but you don't know how many of the scsi drives defects are pending or reallocated, so the (existing) code just adds grown defects as an extra data.  But it has no SMART id because it's not a SMART metric.  There is no right answer for the proper number to go there so I just put -1, because like you said, as far as I know, smart ID's are all positive.  An alternative would be to remove the smart_id label for that one metric, but I leaned toward consistency of form.

Another thing I actually did which was a little more questionable so I did not include it in the PR, was on the actual SATA SMART drives, I also fudge the fudged grown defect count for sata drives to be the sum of reallocated and pending sectors.  I did that so there exists a single metric that represents defects across all (SCSI and SATA) drive types.

dswarbrick:
@GunArm Thanks for the detailed reply. Yep, I also know a little about SMART ;-) https://github.com/dswarbrick/smart

I'll merge this to fix the `lbas_written` ID, and I certainly get your point re: the differences between ATA SMART and SCSI metrics. I think that trying to shoehorn SCSI metrics into a collector that is primarily intended for ATA SMART might be a losing battle. 

There's also the issue of NVMe, which although probably closer to ATA SMART, is still a different animal, and deserves its own collector (which we have in this repo).

PR #128, opened 2022-09-26
Create fstab-check.sh

gnanasalten:
Script to check the filesystems are mounted as per /etc/fstab entry

Signed-off-by: gnanasalten <69344036+gnanasalten@users.noreply.github.com>

### DIFF
--- a/fstab-check.sh
+++ b/fstab-check.sh
@@ -1,12 +1,11 @@
 #!/bin/bash
-path="/var/lib/node_exporter/fstab-check.prom"
 mapfile -t mountpoints < <(awk '$1 !~ /^#/ && $2 ~ /^[/]/ {print $2}' /etc/fstab)
 for mount in "${mountpoints[@]}"
 do
    if ! findmnt "$mount" &> /dev/null
    then
-         echo "node_fstab_mount_status{filesystem=\"$mount\"} 0" | sponge $path
+         echo "node_fstab_mount_status{filesystem=\"$mount\"} 0" >&1
    else
-         echo "node_fstab_mount_status{filesystem=\"$mount\"} 1" | sponge $path
+         echo "node_fstab_mount_status{filesystem=\"$mount\"} 1" >&1
    fi
 done
--- a/fstab-check.sh
+++ b/fstab-check.sh
@@ -1,12 +1,12 @@
 #!/bin/bash
-> /var/lib/node_exporter/fstab-check.prom
-mountpoints=( $(awk '$1 !~ /^#/ && $2 ~ /^[/]/ {print $2}' /etc/fstab) )
+path="/var/lib/node_exporter/fstab-check.prom"
+mapfile -t mountpoints < <(awk '$1 !~ /^#/ && $2 ~ /^[/]/ {print $2}' /etc/fstab)
 for mount in "${mountpoints[@]}"
 do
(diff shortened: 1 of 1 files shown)


dswarbrick:
@gnanasalten Thanks for your contribution.

As you may have noticed, the scripts in this repository output metrics to stdout, to avoid the need to hardcode paths in the script. The recommended way of executing textfile collectors is to capture their stdout with a tool such as `sponge`, as described in the README of this repository.

gnanasalten:
@dswarbrick  thanks for the quick review, I updated the script accordingly, PTAL.

dswarbrick:
@gnanasalten I don't think you got my point.

The script should not call `sponge` itself. It should simply output the metrics to stdout, with the _expectation_ that the end user will run it from a cronjob or systemd timer, piping the output to sponge. This completely avoids the need for your script to know _anything_ about the filesystem layout of the host on which it is running.

gnanasalten:
@dswarbrick updated, PTAL.

fstab-check.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+mapfile -t mountpoints < <(awk '$1 !~ /^#/ && $2 ~ /^[/]/ {print $2}' /etc/fstab)
+for mount in "${mountpoints[@]}"
+do
+   if ! findmnt "$mount" &> /dev/null
+   then
+         echo "node_fstab_mount_status{filesystem=\"$mount\"} 0" >&1
```
dswarbrick, line 7: Redirecting stdout to file descriptor 1 is redundant, since this script does not change the default stdin / stdout / stderr FDs.
```suggestion
         echo "node_fstab_mount_status{filesystem=\"$mount\"} 0"
```

gnanasalten:
@dswarbrick updated, thanks for this quick review.

dswarbrick:
I think that `mountpoint` is a more appropriate choice for the label than `filesystem`... "ext4" is a filesystem.

Please use consistent indentation in the script. I see a mixture of 3-space and 9-space indentation.

gnanasalten:
@dswarbrick sure thing... updated. PTAL.

dswarbrick:
Looks pretty good. The final bits missing are `# HELP` and `# TYPE` lines, as described in the [Prometheus text exposition format](https://github.com/Showmax/prometheus-docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-format-details).

gnanasalten:
@dswarbrick updated... PTAL. , a question is "gauge" correct metric type for this ?

dswarbrick:
> @dswarbrick updated... PTAL. , a question is "gauge" correct metric type for this ?

"gauge" or "untyped" are both valid.

fstab-check.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+echo "# HELP node_fstab_mount_status List and status of filesystem mountpoints"
```
dswarbrick, line 2: This should be fairly obvious to most people, but...
```suggestion
echo "# HELP node_fstab_mount_status List and status of filesystem mountpoints (0 = not mounted, 1 = mounted)"
```

dswarbrick:
LGTM, thanks!

gnanasalten:
@dswarbrick thank you so much for your time!!

PR #134, opened 2023-02-01
nvme_metrics.sh: ensure consistent & predictable numeric format

dswarbrick:
nvme-cli 2.3 and later appears to output numeric values in localized style (even in JSON output), e.g. using comma as thousands separator.

Similar to smartmon.sh, set LC_ALL=C to ensure the least fancy output.

Signed-off-by: Daniel Swarbrick <daniel.swarbrick@gmail.com>

cc: @anarcat

### DIFF
--- a/nvme_metrics.sh
+++ b/nvme_metrics.sh
@@ -9,6 +9,9 @@ set -eu
 #
 # Author: Henk <henk@wearespindle.com>
 
+# Ensure predictable numeric / date formats, etc.
+export LC_ALL=C
+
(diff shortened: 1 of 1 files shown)


anarcat (changes requested):
i typically use C.UTF-8 to avoid things falling over if non-ASCII stuff comes up, just as a general precaution.

anarcat:
oh i didn't know that was a debianism, makes sense then

anarcat (approved):

PR #136, opened 2023-02-06
Add script for ZFS snapshots

lstrojny:
Generates two metrics for ZFS snapshots:
 * zfs_snapshot_space_usage_bytes: snapshot space usage in bytes
 * zfs_snapshot_latest_snapshot_taken: UNIX timestamp

### DIFF
--- a/mock/zfs
+++ b/mock/zfs
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+# zfs mock script for testing zfs snapshot provider
+
+fixtures_dir=$(dirname "$0")/fixtures
+cat "${fixtures_dir}/zfs_list_-p_-H_-t_snapshot_-o_name,used,creation"
--- a/zfs-snapshots.py
+++ b/zfs-snapshots.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+import os
+import subprocess
+from functools import reduce, partial
+from itertools import groupby
+from operator import itemgetter, add
+from prometheus_client import CollectorRegistry, Gauge, generate_latest
+
+
+def row_to_metric(metric: Gauge, row):
+    return metric.labels(pool=row[0][0], volume=row[0][1]).set(row[1])
+
+
+def collect_metrics(metric: Gauge, it) -> None:
+    list(map(partial(row_to_metric, metric), it))
+
+
+def zfs_parse_line(line):
+    cols = line.split("\t")
+    rest, snapshot = cols[0].rsplit("@", 1)
+    pool = rest
+    volume = None
+    if "/" in rest:
+        pool, volume = rest.split("/", 1)
+        volume = "/" + volume
+    return pool, volume, snapshot, *map(int, cols[1:])
+
+
+def zfs_list_snapshots():
+    cmd = [
+        "zfs",
+        "list",
+        "-p",
+        "-H",
+        "-t",
+        "snapshot",
+        "-o",
+        "name,used,creation",
+    ]
+    # zfs list can be relatively slow (couple of seconds)
+    # Use Popen to incrementally read from stdout to not waste further time
+    popen = subprocess.Popen(
+        cmd, stdout=subprocess.PIPE, env=dict(os.environ, LC_ALL="C")
+    )
+    for stdout_line in iter(popen.stdout.readline, ""):
+        stdout_line = stdout_line.strip()
+        if stdout_line == b"":
+            break
+        yield stdout_line.decode("utf-8")
+    return_code = popen.wait()
+    if return_code:
+        raise subprocess.CalledProcessError(return_code, cmd)
+
+
+def aggregate_rows(rows, index, operator):
+    return map(
+        lambda row: (row[0], reduce(operator, map(itemgetter(index), row[1]), 0)), rows
+    )
+
+
+NAMESPACE = "zfs_snapshot"
+LABEL_NAMES = ["pool", "volume"]
+
+
+def main():
+    registry = CollectorRegistry()
+    latest_time_metric = Gauge(
+        "latest_time",
+        "Timestamp of the latest snapshot",
+        labelnames=LABEL_NAMES,
+        namespace=NAMESPACE,
+        registry=registry,
+        unit="seconds",
+    )
+    space_used_metric = Gauge(
+        "space_used",
+        "Space used by snapshots in bytes",
+        labelnames=LABEL_NAMES,
+        namespace=NAMESPACE,
+        registry=registry,
+        unit="bytes",
+    )
+
+    snapshots = map(zfs_parse_line, zfs_list_snapshots())
+    per_fs = list(
+        map(
+            lambda row: (row[0], list(row[1])), groupby(snapshots, lambda row: row[0:2])
+        )
+    )
+
+    space_used = aggregate_rows(per_fs, -2, add)
+    latest_time = aggregate_rows(per_fs, -1, max)
+
+    collect_metrics(latest_time_metric, latest_time)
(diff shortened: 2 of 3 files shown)


dswarbrick:
Please use the official Prometheus client_python library, rather than a homegrown metric printing function.

lstrojny:
@dswarbrick done

zfs-snapshots.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def aggregate_rows(rows, index, operator):
+    return map(
+        lambda row: (row[0], reduce(operator, map(itemgetter(index), row[1]), 0)), rows
+    )
+
+
+NAMESPACE = "zfs_snapshot"
+LABEL_NAMES = ["pool", "volume"]
```
dswarbrick, line 58: It looks like neither of these need to be global vars, as they are only referenced in the `main()` function.

lstrojny: Yup, that's accurate. They are constant values though, that's why I moved them to outer scope

dswarbrick:
@lstrojny Can you also include some sample command outputs (in the `mock/fixtures` directory) and optionally a mock `zfs` script (see existing examples in `mock` directory, so that we can (eventually) write tests for this collector?

zfs-snapshots.py
dswarbrick: Fair enough.

zfs-snapshots.py
```diff
@@ -0,0 +1,96 @@
...
+        "zfs",
+        "list",
+        "-p",
+        "-H",
+        "-t",
+        "snapshot",
+        "-o",
+        "name,used,creation",
+    ]
+    popen = subprocess.Popen(cmd, stdout=subprocess.PIPE)
```
dswarbrick, line 40: Unless you need to interact with the process while it is running (i.e. send input to its `stdin`), I would be more inclined to simply use `subprocess.run()` with `check=True`.

dswarbrick: I would also recommend passing `LC_ALL=C` environment variable to the child process, as we often encounter issues where tools output locale-specific numeric formatting, which can trip up the collector (e.g. European-style comma instead of decimal point or vice versa, dots as thousands separator). Something like `env=dict(os.environ, LC_ALL="C"))` passed to `subprocess.run()` should be sufficient.

mock/zfs
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
dswarbrick, line 1:
```suggestion
#!/usr/bin/env bash
```

zfs-snapshots.py
lstrojny: On `subprocess.run()`: I chose popen because I want to process line by line even if the command hasn’t finished yet.

Passing `LC_ALL` makes tons of sense, changed.

dswarbrick: > On `subprocess.run()`: I chose popen because I want to process line by line even if the command hasn’t finished yet.

Why is that important? The resulting .prom file (e.g. when using `sponge`) is not going to be written until the collector script finishes anyway. Collectors should not output "partially successful" metrics - it should be all or nothing (and exit with an error).

mock/fixtures/zfs list -p -H -t snapshot -o name,used,creation
dswarbrick, line 1: Please avoid spaces in the filename. Use underscores instead.

zfs-snapshots.py
lstrojny: It’s not about the output of the collector but about the output of the zfs command. It’s relatively slow (takes on a system with 11K snapshots around 5 seconds) so I am trying to shave off every millisecond I get so processing as soon as any output arrives from the zfs command is to not waste any time in an otherwise already slowish command.

dswarbrick: Ok, fair enough. Perhaps add a code comment then, explaining the motivation behind it, so that it doesn't merely look like unnecessary over-engineering.

mock/fixtures/zfs list -p -H -t snapshot -o name,used,creation
lstrojny: Fixed

zfs-snapshots.py
lstrojny: Done

zfs-snapshots.py
```diff
@@ -0,0 +1,100 @@
...
+        map(
+            lambda row: (row[0], list(row[1])), groupby(snapshots, lambda row: row[0:2])
+        )
+    )
+
+    space_usage = aggregate_rows(per_fs, -2, add)
+    latest_snapshots = aggregate_rows(per_fs, -1, max)
+
+    collect_metrics(latest_snapshot_taken_metric, latest_snapshots)
+    collect_metrics(space_usage_metric, space_usage)
```
dswarbrick, line 94: Return value of `collect_metrics` function is ignored. Is this intentional?

zfs-snapshots.py
```diff
@@ -0,0 +1,100 @@
...
+    registry = CollectorRegistry()
+    latest_snapshot_taken_metric = Gauge(
+        "latest_snapshot_taken",
+        "Timestamp of the latest snapshot",
+        labelnames=LABEL_NAMES,
+        namespace=NAMESPACE,
+        registry=registry,
+    )
+    space_usage_metric = Gauge(
+        "space_usage",
```
dswarbrick, line 75: nit: `foo_used` is more widely accepted naming convention
```suggestion
        "space_used",
```

zfs-snapshots.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+NAMESPACE = "zfs_snapshot"
+LABEL_NAMES = ["pool", "volume"]
+
+
+def main():
+    registry = CollectorRegistry()
+    latest_snapshot_taken_metric = Gauge(
+        "latest_snapshot_taken",
```
dswarbrick, line 68: Looking at similar epoch / timestamp style metrics in libraries like prometheus/common, e.g.:
```
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.68509617503e+09
```
I would be inclined to name this `latest_time_seconds` (since "snapshot" is already in the metric prefix).

zfs-snapshots.py
lstrojny: Done

zfs-snapshots.py
lstrojny: Added type hints to indicate that collect_metrics is a void func

zfs-snapshots.py
lstrojny: Changed this one as well to include the unit. I am wondering if this should rather be modelled as snapshot age, meaning number of seconds since the last snapshots. Any opinion?

zfs-snapshots.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+NAMESPACE = "zfs_snapshot"
+LABEL_NAMES = ["pool", "volume"]
+
+
+def main():
+    registry = CollectorRegistry()
+    latest_snapshot_taken_metric = Gauge(
+        "latest",
```
dswarbrick, line 68: nit: Just to make it clear that it's a time(stamp), not a duration.
```suggestion
        "latest_time",
```

dswarbrick (approved):

dswarbrick:
Great stuff, thanks for your contribution.

PS: Any interest in merging the functionality of zfs_zpool.sh into this collector? :wink:

lstrojny:
Thanks! I probably won’t find the time to move zfs_zpool.sh into this collector, although a general purpose zfs.py would be nice

PR #142, opened 2023-04-04
apt_info.py: Use official Prometheus Python client library

dswarbrick:
Replace homegrown text exposition format rendering with official Prometheus Python client library registry and functions.


dswarbrick:
poke @SuperQ

SuperQ (approved):
Nice

PR #145, opened 2023-05-21
Migrate CircleCI config to GitHub Actions

dswarbrick:
Implement Python flake8 linting and shellcheck as GitHub Actions. Preserve stub CircleCI config with no-op workflow.


dswarbrick:
shellcheck `SC2086` (double quote to prevent globbing and word splitting) excluded for now, since two scripts in the repo fail that check, which were apparently not being caught by the previous CircleCI workflow. I will fix those scripts in a separate PR and reinstate the `SC2086` check then.

anarcat (approved):
lgtm.

SuperQ (approved):

PR #152, opened 2023-05-22
Remove smartmon run metric

SuperQ:
This metric is a bit funny, it's a timestamp per device, but it's not accurate enough (seconds, not milliseconds) to derive much useful information from it. This should probably be replaced with a per-device duration metric.

### DIFF
--- a/smartmon.py
+++ b/smartmon.py
@@ -2,7 +2,6 @@
 import argparse
 import collections
 import csv
-import datetime
 import decimal
 import re
 import shlex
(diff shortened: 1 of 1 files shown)


dswarbrick:
Is this related to my necropost on #104? ;-)

Personally I don't know what the point of this metric is, and the help text is equally vague. If the original author wanted timestamps of when the metrics were actually collected for the device, this is not the right way to go about it.

I suspect they were unaware that the Prometheus text exposition format allows optional timestamps as the last field on the line.

SuperQ:
Yes, but we forbid timestamps on textfile collector metrics. And there's already a node_exporter mtime metric for when the last textfile was updated.

dswarbrick:
I was not aware that timestamps are forbidden on textfile collector metrics. I suspect I probably attempted to use them on the odd occasion in the past.

In any case, I don't see that this `smartctl_run` metric adds anything of significant value that cannot be derived from `node_textfile_mtime_seconds`, and I agree that a per-device collection duration would be more useful.

dswarbrick (approved):

PR #155, opened 2023-05-23
Add new Python-based nvme_metrics collector

dswarbrick:
This is a Python reimplementation of the existing nvme_metrics.sh script. It features several new metrics:
- nvme_device_info (model, firmware, serial number)
- nvme_physical_size_bytes
- nvme_used_bytes
- nvme_sector_size_bytes
- nvme_unsafe_shutdowns_total

It retains backwards compatibility with the metric names of nvme_metrics.sh with one exception:
The original nvme_critical_warning_total metric is incorrect, because the source is not a counter. The NVMe critical warning field, defined in section 5.16.1.3 of the NVMe Base Specification 2.0c, is a bitmap, and thus should be a gauge in Prometheus. As such, the metric is renamed to simply nvme_critical_warning.

It is my intention that this script should (eventually) supersede and replace the existing `nvme_metrics.sh` script.

### DIFF
--- a/nvme_metrics.py
+++ b/nvme_metrics.py
@@ -41,12 +41,12 @@ metrics = {
     ),
     "data_units_read": Counter(
         "data_units_read_total",
-        "Device data units read",
+        "Number of 512-byte data units read by host, reported in thousands",
         ["device"], namespace=namespace, registry=registry,
     ),
     "data_units_written": Counter(
         "data_units_written_total",
-        "Device data units written",
+        "Number of 512-byte data units written by host, reported in thousands",
         ["device"], namespace=namespace, registry=registry,
     ),
     "device_info": Gauge(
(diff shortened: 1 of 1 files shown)


nvme_metrics.py
```diff
@@ -0,0 +1,222 @@
...
+# Disable automatic addition of _created series. Must be set before importing prometheus_client.
+os.environ["PROMETHEUS_DISABLE_CREATED_SERIES"] = "true"
+
+from prometheus_client import CollectorRegistry, Counter, Gauge, generate_latest  # noqa: E402
+
+registry = CollectorRegistry()
+namespace = "nvme"
+
+metrics = {
+    # fmt: off
```
SuperQ, line 26: :joy:

SuperQ (approved):
Nice

PR #159, opened 2023-06-06
add support for monitoring LVM caches

anarcat:
This is basically a copy-paste of the snapshots code, and seems to work on my end.

Closes: #158

PR #160, opened 2023-06-06
lvm: add support for "all" checks

anarcat:
It seems rather inconvenient to have to specify all those switches by hand. Right now, I need to specify `-gpst` to get all the stats and, if some new metric get added (e.g. #159 hopefully), I need to remember to update my callers to match. This is annoying.

Let's make a `-a` that's going to do it all for us. Hell, i'd make it the default, personally, if I wasn't worried about breaking people's stuff...

This obviously conflicts with other PRs that might add another flag like this, like #159. Please prioritize the latter.

### DIFF
--- a/lvm-prom-collector
+++ b/lvm-prom-collector
@@ -48,8 +48,14 @@ snapshots=false
 physical=false
 groups=false
 
-while getopts "htpsg" opt; do
+while getopts "ahtpsg" opt; do
   case $opt in
+    a)
+      thin_pools=true
+      snapshots=true
+      physical=true
+      groups=true
+      ;;
     p)
       physical=true
       ;;


SuperQ (approved):
LGTM.

The only issue with `-a` as default is there's no `false` switch on these. So there would be no way to turn them off.

SuperQ:
Oh, oops, can you followup to this by adding `-a` to the usage text?

anarcat:
> The only issue with `-a` as default is there's no `false` switch on these. So there would be no way to turn them off.

i guess that the default would be all unless one of the flags is specified, in which case only the ones specific trigger...

anarcat:
> Oh, oops, can you followup to this by adding `-a` to the usage text?

fixed in #161

PR #162, opened 2023-06-09
Script to create text file for hp ssacli tool output

epwickey:
This is a script for creating text file with output from hp ssacli tool to monitor hp raids

### DIFF
--- a/hpssacli.sh
+++ b/hpssacli.sh
@@ -36,15 +36,15 @@ parse_ssacli_controller(){
     local controller_mats=("$@")
     for matric in "${controller_mats[@]}"
     do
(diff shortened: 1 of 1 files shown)


dswarbrick:
Are you aware that there is a dedicated https://github.com/sahandhabibi/Ssacli-Exporter?

epwickey:
> Are you aware that there is a dedicated https://github.com/sahandhabibi/Ssacli-Exporter?

I am. However since I had to create this since we already use node-exporter I thought I will contribute. Also the exporter only works for servers with one slot and only export physical disk metrics while this script can collect metrics for servers with multiple slots and it also collects metrics about the virtual disks and controller.

dswarbrick:
The shellcheck errors will need to be fixed before this can be considered.

PR #163, opened 2023-06-16
fix storcli exporter behavior when storcli is not executable

avk999:
SystemExit should be raised, not called


dswarbrick (approved):
Thanks @avk999. This always looked kinda strange to me, but I never got around to investigating.

PR #164, opened 2023-06-29
add "4096" as healthy BBU code

avk999:
Old PERC controllers (at least PERC H730/740 Mini) with FW 4.270.00-8168 return BBU status 4096. Comparing the output with MegaCLI output it seems that this bit corresponds to "Learning cycle required" indicator. However since controller runs learning cycles by itself there is no reason for alerting.


dswarbrick:
I have a feeling that as more of these values are discovered, this collector will have to change from a `foo in [0, 8, 4096]` test to actually testing the bits of the value with a bitwise AND. The current naive approach could potentially break if / when other bits in this value are set.

dswarbrick (approved):

PR #166, opened 2023-07-04
Remove available/spare/user percentage calculation

knowhy:
This PR removes incorrect calculations for some NVMe related metrics.

According to the [NVM Express Base Specification](https://nvmexpress.org/wp-content/uploads/NVM-Express-Base-Specification-2.0c-2022.10.04-Ratified.pdf) the fields `Available Spare`, `Available Spare Threshold` and `Percentage Used` return normalized percentage values.

So the dividing the value by `100` is not necessary and returns incorrect percentage values.

### DIFF
--- a/nvme_metrics.py
+++ b/nvme_metrics.py
@@ -183,9 +183,9 @@ def main():
         metrics["host_write_commands"].labels(device_name).inc(
             int(smart_log["host_write_commands"])
         )
-        metrics["avail_spare"].labels(device_name).set(smart_log["avail_spare"] / 100)
-        metrics["spare_thresh"].labels(device_name).set(smart_log["spare_thresh"] / 100)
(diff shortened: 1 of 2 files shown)


dswarbrick:
The divide by 100 is deliberate, because the percentage values returned by `nvme-cli` are converted to _ratios_ (0 - 1.0), e.g.:
```python
    "spare_thresh": Gauge(
        "available_spare_threshold_ratio",
        "Device available spare threshold ratio",
        ["device"], namespace=namespace, registry=registry,
    ),
```
This is in compliance with the [best practices](https://prometheus.io/docs/practices/naming/#base-units) for expressing percentages as Prometheus metrics.

knowhy:
Ok. Sorry for the noise. I got confused by the values over 100%.

PR #171, opened 2023-07-31
Support host prefix for apt_info.py

lukasmrtvy:
This will allow us to run apt_info.py on the host from the container.

```
docker run --rm -it --entrypoint bash -v "/:/hostfs" ubuntu:22.04 -c 'apt update && apt install -y python3-prometheus-client python3-apt curl && curl https://raw.githubusercontent.com/prometheus-community/node-exporter-textfile-collector-scripts/7aaf51790bf19911ebcaa49f6474b26bd04b3910/apt_info.py -o /tmp/apt_info.py && APT_HOSTPREFIX=/hostfs/ python3 /tmp/apt_info.py'
```

### DIFF
--- a/apt_info.py
+++ b/apt_info.py
@@ -111,7 +111,7 @@ def _write_cache_timestamps(registry):
 def _write_reboot_required(registry):
     g = Gauge('node_reboot_required', "Node reboot is required for software updates.",
               registry=registry)
-    g.set(int(os.path.isfile(os.path.join(hostPrefix,'/run/reboot-required'))))
+    g.set(int(os.path.isfile(os.path.join(hostPrefix,'run/reboot-required'))))
 
 
 def _main():


dswarbrick:
This looks to me like more of a job for [chroot](https://man7.org/linux/man-pages/man1/chroot.1.html).

lukasmrtvy:
@dswarbrick Why? python3-prometheus-client python3-apt dependencies are installed only in the container

dswarbrick:
What advantage does running this script in Docker actually bring you? It seems like an unnecessarily complicated setup.

Assuming it's a Debian or Ubuntu host, these scripts are packaged as [prometheus-node-exporter-collectors](https://packages.debian.org/search?keywords=prometheus-node-exporter-collectors).

lukasmrtvy:
Ok, I will try a different approach... There are people ( https://unix.stackexchange.com/questions/261165/change-cache-location-path-for-apt )  having apt dir elsewhere than in the default path, this PR will allow supporting different apt-cache paths.

dswarbrick:
I can understand the motivation to support a non-standard apt cache location, and would accept that being implemented via a command-line option.

I still don't understand your motivation to run this script in a container, considering the hoops that you have to jump through in order to make it work. It is designed to operate on the _host's_ apt cache, and can run as an unprivileged user. Why do you feel the need to sandbox it in a container?

But you also want to override the path to `/run` - in other words, you want to prefix all paths that this script attempts to open. And there is a tried and tested method for doing that, called chroot. Let's not reinvent the wheel here.

egvimo:
I would also like to see this feature implemented.

I'm running my node exporter via Kube Prometheus Stack directly inside Kubernetes. Including all other monitoring related stuff. So my nodes already have the the root file system mounted to other node exporter containers.

Therefore I don't want to do a two-pronged approach and install this script directly on the node.

anarcat:
this PR should be closed.

PR #172, opened 2023-08-25
Fix typo in readme file

saeedhosseini21:
Fix typo in readme file


dswarbrick:
That is not a typo. It is completely intentional.

SuperQ:
Atomic is the correct term.

PR #173, opened 2023-08-26
Systemd unit and timer for smartmon

LukeLR:
I added systemd unit and timers for smartmon.sh, so they can be easily installed for running the smartmon script periodically as a service. What do you think? :)

### DIFF
--- a/smartmon.service
+++ b/smartmon.service
@@ -2,7 +2,7 @@
 Description=Gather S.M.A.R.T. metrics to be picked up by prometheus node exporter
 
 [Service]
-ExecStart=smartmon.sh | sponge /tmp/smartmon.prom
+ExecStart=bash -c "smartmon.sh | sponge /tmp/smartmon.prom"
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


dswarbrick:
Generally we leave such things to the distros which package the contents of this repo, since it usually involves distro-specific paths and users. For example, the Debian (and Ubuntu) package [prometheus-node-exporter-collectors](https://packages.debian.org/search?keywords=prometheus-node-exporter-collectors) already includes systemd units and timers for five of the scripts in this repo.

LukeLR:
Thanks for the reply. That's unfortunate, especially since no distro-specific paths are required if the scripts are simply installed in $PATH. I'd like to see sytemd units upstream, even if distros package systemd units, they can use them as an example. But if you don't want that in, I'll maintain a fork for myself.

dswarbrick:
I think that if we're going to include such things in this repo, they belong in a "contrib" directory, rather than the root directory. You'll also need to sign the DCO if you wish these to be considered for merging.

PR #175, opened 2023-09-02
Some disks doesnt report device speed correctly, crashing script execution

ARPABoy:
Some disks doesnt report device speed correctly when pd_device_speed metric assigns a string value: "Unknown" script execution crashes in that case assign a 0 speed value.


dswarbrick:
My gut feeling is that transforming an "unknown" value to a value of zero is going to obscure cases where disks are spun-down, and truly have a rotational speed of zero. I think it would be better to simply omit the metric if it is not a valid numeric value (which can be tested with `isnumeric()`).

Issue #119, opened 2022-05-19
Don't maintain two smartmon scripts


varac:
I'm confused by the current state that this repo maintains two smartmon scripts (one shell, one pythons script).
I'd like to propose to decide on one that should be maintained, and remove the other, to bundle the efforts and also not confuse users (which one should I use ? what are the differences ?).
Debian packages (still) use the shell script by default. The last commit to the shell script was 2 years ago, the last commit for the python script 1 year ago.

iustin:
In the meantime, there have been a few more commits to the shell one. And the python one is not using proper exporter library to generate the output, but instead does manual `print()` statments. And neither the shell one nor the python one are using the json output from smartctl, relying instead on parsing the text output :(

It looks like it would be time to write a new python (or whatever) script from scratch, that is more clean/modern.

dswarbrick:
> And neither the shell one nor the python one are using the json output from smartctl, relying instead on parsing the text output :(

IIRC, that was due to the fact that smartctl only supports JSON output from v7.x onwards (again, IIRC).

I would definitely be in favour of _requiring_ the use of official Prometheus python library for all the collector scripts here however. Too often I see homebaked implementations choke on quoting or non-ASCII chars, and the quality of implementation is quite variable.

dswarbrick:
This repo doesn't really have "versions" per se. We could move the legacy collectors to an "attic" directory perhaps, which distros could omit from their packaged releases (e.g. Debian's [prometheus-node-exporter-collectors](https://packages.debian.org/search?keywords=prometheus-node-exporter-collectors)). Being that this is a git repo though, with full version history, I'm not sure if even that would be necessary. If somebody wants the old collectors, they can use an older git snapshot. I doubt that any distros are still publishing new releases with pre-7.x smartmontools.

The main requirement is that a new / replacement collector did not break backwards compatibility in terms of metric names and labels. Unit tests would also be really nice, i.e. include some JSON test fixtures and write the collector in a modular way such that it can be tested against them.

bjornfor:
After learning that smartmon.py doesn't expose SMART ID 11 (Calibration Retry Count), I migrated to https://github.com/prometheus-community/smartctl_exporter. (That project already parses JSON data from smartctl.)

Why not deprecated both smartmon.{sh,py} and suggest using smartctl_exporter?

cpatulea:
December 2024 update:
- I was also confused by the two scripts and spent a while chasing ghosts
[...]

PR #178, opened 2023-09-18
Allow setting different root dir for apt

egvimo:
See #171 

Implemented as command line argument instead of env var.

Additionally I've refactored a little bit. Fixed linting issues and created a separate function to use in other scripts.

### DIFF
--- a/apt_info.py
+++ b/apt_info.py
@@ -20,10 +20,11 @@
 # Authors: Kyle Fazzari <kyrofa@ubuntu.com>
 #          Daniel Swarbrick <dswarbrick@debian.org>
 
-import apt
-import apt_pkg
+import argparse
 import collections
 import os
+import apt
+import apt_pkg
 from prometheus_client import CollectorRegistry, Gauge, generate_latest
 
 _UpgradeInfo = collections.namedtuple("_UpgradeInfo", ["labels", "count"])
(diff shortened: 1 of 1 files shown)


macetw:
Definitely needed to manage within a docker container. Suggest moving this forward.

macetw (approved):

macetw:
This PR is only managing it for the apt script. Do we need this setting for any of the other scripts in this code?

egvimo:
I've created an exporter using the script from this PR which runs inside a Kubernetes cluster inside a container exporting the metrics via `ServiceMonitor`: https://github.com/egvimo/apt-exporter
This works even with a read only mount of the host directory.

anarcat:
should this be merged, or should we look at removing apt_info.py in favor of the apt-exporter?

dswarbrick:
> should this be merged, or should we look at removing apt_info.py in favor of the apt-exporter?

If we remove a script simply because a dedicated exporter exists, then we would also need to remove the smartmon.{py,sh} scripts.

anarcat:
> If we remove a script simply because a dedicated exporter exists, then we would also need to remove the smartmon.{py,sh} and ipmitool scripts.

that would solve #119 at least :p

Issue #179, opened 2023-10-11
apt_info.py can hang for hours


anarcat:
We have a situation here where numerous machines are seeing the `apt_info.py` script hang for hours. This has been reported in Debian as bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1028212 and is being tracked in our internal tracker as https://gitlab.torproject.org/tpo/tpa/team/-/issues/41355

This is possibly an upstream issue, part of python-apt, requires further investigation. Current workaround includes installing a `TimeoutStartSec=30s` in the systemd unit `[Service]` block.

anarcat:
In https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1028212#62, @julian-klode has an excellent point that `apt_info.py` should *not* trigger `apt update` on its own. There are out of band mechanisms to do this (e.g. unattended-upgrades or apt itself, actually asked jak for clarification on that) that are supposed to take care of this. `apt_info.py`'s job is simply to collect the numbers and dump them into something prometheus can parse. It should do the least possible for this.

For out of date mirrors or data, it should report the cache age (#180), not try to pre-emptively update it.

So i'll make a PR to fix that shortly.

PR #181, opened 2023-10-13
do not run apt update or simulate apt dist-upgrade

anarcat:
This is causing all sorts of problems. The first of which is that we're hitting our poor mirrors every time the script is ran, which, in the Debian package configuration, is *every 15 minutes* (!!).

The second is that this locks the cache and makes this script needlessly stumble upon a possible regression in APT from Debian bookworm and Ubuntu 22.06:

https://bugs.launchpad.net/ubuntu/+source/apt/+bug/2003851

That still has to be confirmed: it's possible that `apt update` can hang for a long time, but that shouldn't concern us if we delegate this work out of band.

I also do not believe actually performing the `dist-upgrade` calculations is necessary to compute the pending upgrades at all. I've done work with python-apt for other projects and haven't found that to be required: the cache has the necessary information about pending upgrades.

Closes: #179

### DIFF
--- a/apt_info.py
+++ b/apt_info.py
@@ -3,6 +3,18 @@
 # Description: Expose metrics from apt. This is inspired by and
 # intended to be a replacement for the original apt.sh.
 #
+# This script deliberately does *not* update the apt cache. You need
+# something else to run `apt update` regularly for the metrics to be
+# up to date. This can be done in numerous ways, but the canonical way
+# is to use the normal `APT::Periodic::Update-Package-Lists`
+# setting.
+#
+# This, for example, will enable a nightly job that runs `apt update`:
+#
+#     echo 'APT::Periodic::Update-Package-Lists "1";' > /etc/apt/apt.conf.d/99_auto_apt_update.conf
+#
+# See /usr/lib/apt/apt.systemd.daily for details.
+#
 # Dependencies: python3-apt, python3-prometheus-client
 #
 # Authors: Kyle Fazzari <kyrofa@ubuntu.com>
@@ -10,7 +22,6 @@
 
 import apt
 import collections
-import contextlib
 import os
 from prometheus_client import CollectorRegistry, Gauge, generate_latest
 
@@ -82,15 +93,6 @@ def _write_reboot_required(registry):
 
 def _main():
     cache = apt.cache.Cache()
-
-    # First of all, attempt to update the index. If we don't have permission
-    # to do so (or it fails for some reason), it's not the end of the world,
-    # we'll operate on the old index.
-    with contextlib.suppress(apt.cache.LockFailedException, apt.cache.FetchFailedException):
-        cache.update()
-
-    cache.open()
-    cache.upgrade(True)
 
     registry = CollectorRegistry()
     _write_pending_upgrades(registry, cache)


dswarbrick:
I was never a fan of this script updating the cache, and suggested on at least one occasion that we'd be better off running the script as an unprivileged user by default, just so that it _wouldn't_ try to update the cache.

OTOH, there will be people who run this script _without_ asynchronous updating of the cache (via e.g. apticron or unattended-upgrades). I'd sorta prefer to put this cache-updating code behind a command-line switch (disabled by default), for such cases.

anarcat:
On 2023-10-13 09:56:29, Daniel Swarbrick wrote:
> OTOH, there will be people who run this script _without_ asynchronous updating of the cache (via e.g. apticron or unattended-upgrades). I'd sorta prefer to put this cache-updating code behind a command-line switch (disabled by default), for such cases.

The problem with this is that it's a policy decision that's largely
disconnected from when you want to report on the actual state of the
system.

In the Debian package, for example, this script runs every fifteen
minutes. If you run apt-update every fifteen minutes on all your
machines, you are going to hurt the mirrors, and you're being a bad
citizen. You shouldn't do that.

We were doing this, but we didn't even *know* we were doing this. We
were just running this small metric here, surely that causes no harm?

I *really* don't think this should run apt-update. It's slow, and it's
hurting the mirrors, it's just a plain bad idea.

If people need a job to update the cache, they should make a job to
update the cache, i don't get why it would be the exporter's job to do
so.

The previous shell implementation of this, by the way, did not run apt
update either, so I consider this a serious regression of the apt update
rewrite.

I'm not completely closed to rewriting this to have a commandline flag
to enable apt update, but I think it's overkill, and I am not
particularly interested in working on this myself.

Finally, keep in mind this patch is aimed for inclusion in a Debian
stable release. The more complex we make it, the more trouble it will be
to ship there (and elsewhere).

dswarbrick:
I don't see that it would be too complex to simply ensure that the cache update is not performed unless explicitly opted into (for those who need it). By making the flag disabled by default, no extra changes to the .service would be needed. It would merely be a behaviour change (albeit one that could be reverted, unlike stripping out the functionality entirely).

This repo has no tags, so is effectively a rolling release. Therefore what Debian stable or any other distro which packages these scripts does is up to that distro.

anarcat:
On 2023-10-13 11:31:43, Daniel Swarbrick wrote:

[...]

> This repo has no tags, so is effectively a rolling release. Therefore what Debian stable or any other distro which packages these scripts does is up to that distro.

I am a Debian developer. I am providing this very patch as a hotfix for
this problem in Debian specifically. If you want to make it more
elaborate, I guess it's your call, but I would suggest taking the patch
as is and adding the extra feature on top after, that way we can get
moving here... :)

Are we in agreement that, by default, we shouldn't run apt-get update
here? This is what this patch does...

anarcat:
Also, I did a bit of research and the normal way of ensuring the package list is up to date on APT systems is to enable the `APT::Periodic::Update-Package-Lists` setting in `apt.conf`. That's as simple as:

    echo 'APT::Periodic::Update-Package-Lists "1";' > /etc/apt/apt.conf.d/99_auto_apt_update.conf

I hardly see why this specific script should do that job instead...

anarcat:
I think one key concept that I might not have mentioned here is that this script reports on two different tasks: updating the package list and actually upgrading pending packages. The latter can be done pretty much any time an admin passes by a server or unattended-upgrades runs, and it's worth running often to check for that. The former, however, is needless to run repeatedly because mirrors themselves (heck, ftp-master.debian.org even) do not upgrade that often (dinstall runs every 15 minutes if my memory is correct).

So we really shouldn't do any action here: just like this script doesn't (and shouldn't!) perform actual upgrades, I don't think it should update the package list either. I really like the idea of having the script deliberately readonly, and I think it's going to make everything simpler if we keep it like that.

But again, if you want to take this further, feel free... I just don't want to get bogged down in those implementation details right now.

SuperQ:
Yes, I would prefer users update the apt-cache on their own schedule and this script does not.

SuperQ (approved):
Would you mind adding the apt config notes as a documentation comment to the top of the script?

anarcat:
> Would you mind adding the apt config notes as a documentation comment to the top of the script?

done.

SuperQ (approved):

apt_info.py
```diff
@@ -3,14 +3,25 @@
 # Description: Expose metrics from apt. This is inspired by and
 # intended to be a replacement for the original apt.sh.
 #
+# This script does *not* update the apt cache, deliberately. You need
```
dswarbrick, line 6:
```suggestion
# This script deliberately does *not* update the apt cache. You need
```

apt_info.py
```diff
@@ -3,14 +3,25 @@
...
+# something else to run `apt update` regularly for the metrics to be
+# up to date. This can be done un numerous ways, but the canonical way
```
dswarbrick, line 8:
```suggestion
# up to date. This can be done in numerous ways, but the canonical way
```

dswarbrick (changes requested):

anarcat:
thanks for the review and (tacit?) approval, @dswarbrick  :) i integrated your changes in the commit with credit in the commitlog.

dswarbrick (approved):

anarcat:
i think this is now ready to merge, @SuperQ can you push the magic button? :)

Issue #180, opened 2023-10-11
apt info should report cache age


anarcat:
We cannot alert on stale apt caches. Combined with #179, we can quickly end up in a situation where critical security upgrades are delayed or never installed.

We should have a timestamp metric showing when the last apt update was ran. Optionnally, we could have that metric per mirror as well.

anarcat:
@julian-klode what would be the canonical way to check when `apt update` ran last? maybe the last mod time on `/var/cache/apt/pkgcache.bin`? i guess we could also populate a separate metric based on the `/var/lib/apt/lists/*InRelease` timestamp, or should we parse the `Date:` field in those?? 

thanks, and sorry for the ping if that's inappropriate...

anarcat:
i picked only `pkgcache.bin` in #182 for now, with the understanding that it's an important metric to add once we remove the auto-update in #181. This could be extended to cover mirror ages later on, but I suspect that would trigger a lot of noise in a larger fleet, as each server would individually report that its list is out of date... I guess an alerting rule could be tweaked to avoid alerting too often, but still... seems overkill for now.

geor-g:
`apt` ships [apt.systemd.daily](https://salsa.debian.org/apt-team/apt/-/blob/main/debian/apt.systemd.daily#L82), which gets installed into `/usr/lib/apt/`. It provides `check_stamp()` and relies on `/var/lib/apt/periodic/update-stamp`. I can't tell if it makes sense to rely on this code, but in any case this might provide an idea how to implement this.

anarcat:
yeah, in #181 i actually document how that works too, or at least an example of it.

PR #182, opened 2023-10-13
report the apt cache timestamp

anarcat:
We use the `pkgcache.bin` modification time as a heuristic, but there might be a better way to do this.

We also silently ignore errors when pulling that file to avoid broken systems from breaking the other metrics. I believe this will lead to a null metric which is probably what we want anyway.

A possible improvement to this would be to individually inspect the `InRelease` files and report the mirror's timestamps as well, but that's more involved. It's also not a priority compared to this fix, because we want the update timestamp if we remove the `apt update` run (in #181).

Closes: #180

### DIFF
--- a/apt_info.py
+++ b/apt_info.py
@@ -74,6 +74,14 @@ def _write_autoremove_pending(registry, cache):
     g.set(len(autoremovable_packages))
 
 
+def _write_cache_timestamps(registry):
+    g = Gauge('apt_package_cache_timestamp_seconds', "Apt update last run time.", registry=registry)
+    try:
+        g.set(os.stat('/var/cache/apt/pkgcache.bin').st_mtime)
+    except OSError:
+        pass
+
+
 def _write_reboot_required(registry):
     g = Gauge('node_reboot_required', "Node reboot is required for software updates.",
               registry=registry)
@@ -96,6 +104,7 @@ def _main():
     _write_pending_upgrades(registry, cache)
     _write_held_upgrades(registry, cache)
     _write_autoremove_pending(registry, cache)
+    _write_cache_timestamps(registry)
     _write_reboot_required(registry)
     print(generate_latest(registry).decode(), end='')
 


apt_info.py
```diff
@@ -74,6 +74,14 @@ def _write_autoremove_pending(registry, cache):
     g.set(len(autoremovable_packages))
 
 
+def _write_cache_timestamps(registry):
+    g = Gauge('apt_update_timestamp_seconds', "Apt update last run time.", registry=registry)
```
SuperQ, line 78:
```suggestion
    g = Gauge('apt_package_cache_timestamp_seconds', "Apt update last run time.", registry=registry)
```

SuperQ (approved):
Minor naming nit, otherwise, LGTM.

anarcat:
merged your suggestion in the commit, along with a credit in commitlog, should be ready to merge now.

SuperQ (approved):

julian-klode:
This has all a bit too fast for me to actually give any feedback (I'm afk all week except this morning more or less), but pkgcache.bin will also get updated whenever you install or upgrade or remove a package. `srcpkgcache.bin` is rebuilt whenever the sources.list changes, you ran clean, or the lists actually changed.

The best approach arguably is to check `/var/lib/apt/periodic/update-success-stamp` to see when the last *succesful* update was - but it is only set by the periodic script, so that only makes sense if the option is set (querying that with `apt_pkg.init_config(); apt_pkg.config["APT::Periodic::Update-Package-Lists"]` is nicest if you want to use `python3-apt`, but could do `apt-config get` too).

The best proxy otherwise is just the mtime of the `/var/lib/apt/lists` directory as we always `rename()` files from the partial directory into there, so it should always be updated as far as I understand directory modification times. The modification times of the files meanwhile are the modification times on the server and hence not meaningful.

This is a common issue that other tools also have and APT itself will need to deal with, I think we're going to end up adding actual stamp files to the apt code itself so you can always tell when the last update was and whether there were errors. Maybe just dump a (machine-readable) update.log in there, once we have machine-readable error codes.

Issue #183, opened 2023-10-17
apt cache age computed incorrectly


anarcat:
in #182, @julian-klode said:

> This has all a bit too fast for me to actually give any feedback (I'm afk all week except this morning more or less), but pkgcache.bin will also get updated whenever you install or upgrade or remove a package. `srcpkgcache.bin` is rebuilt whenever the sources.list changes, you ran clean, or the lists actually changed.
>
> The best approach arguably is to check `/var/lib/apt/periodic/update-success-stamp` to see when the last *succesful* update was - but it is only set by the periodic script, so that only makes sense if the option is set (querying that with `apt_pkg.init_config(); apt_pkg.config["APT::Periodic::Update-Package-Lists"]` is nicest if you want to use `python3-apt`, but could do `apt-config get` too).
>
> The best proxy otherwise is just the mtime of the `/var/lib/apt/lists` directory as we always `rename()` files from the partial directory into there, so it should always be updated as far as I understand directory modification times. The modification times of the files meanwhile are the modification times on the server and hence not meaningful.
>
> This is a common issue that other tools also have and APT itself will need to deal with, I think we're going to end up adding actual stamp files to the apt code itself so you can always tell when the last update was and whether there were errors. Maybe just dump a (machine-readable) update.log in there, once we have machine-readable error codes.

_Originally posted by @julian-klode in https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/issues/182#issuecomment-1765726265_

So we need to tweak our cache age metric to check the mtime on `/var/lib/apt/lists` until APT standardizes this. I don't feel like relying on `APT::Periodic` stuff is sufficiently flexible to cover most use cases (few people seem aware of that feature in the first place...)

Looking at mirror files is still interesting, I think. Reporting on out-of-date mirrors is a worthwhile, but separate goal, so let's not get bogged down on that here.

PR #184, opened 2023-10-17
use a better heuristic for the apt update last run time

anarcat:
As reported in #182, `pkgcache.bin` gets updated on more than just `apt update`. In particular, it gets modified when a package is installed, upgraded or removed.

So let's fallback on a better heuristic, which is the `APT::Periodic` timestamp. This should give us a much more precise and deliberate status.

We also fallback to the `lists` directory, which gets updated when new mirror lists gets moved into place. This does run the risk of staying silently unchanged if there's no change on mirrors but (a) that's rather infrequent and (b) we might actually want to warn on that anyway.

Closes: #183

### DIFF
--- a/apt_info.py
+++ b/apt_info.py
@@ -21,6 +21,7 @@
 #          Daniel Swarbrick <dswarbrick@debian.org>
 
 import apt
+import apt_pkg
 import collections
 import os
 from prometheus_client import CollectorRegistry, Gauge, generate_latest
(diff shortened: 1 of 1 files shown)


anarcat:
@julian-klode could you review this? I think it covers the comments you made in #182...

jcharaoui:
If `APT::Periodic` is used at some point, and then disabled, wouldn't this report a false metric  if `apt` is updated in another manner?

anarcat:
On 2023-10-17 07:50:27, Jerome Charaoui wrote:
> If `APT::Periodic` is used at some point, and then disabled, wouldn't this report a false metric  if `apt` is updated in another manner?

If it's disabled, the code won't check that timestamp file, unless I got
this wrong.

jcharaoui:
> On 2023-10-17 07:50:27, Jerome Charaoui wrote: If `APT::Periodic` is used at some point, and then disabled, wouldn't this report a false metric if `apt` is updated in another manner?
> If it's disabled, the code won't check that timestamp file, unless I got this wrong.

Right, my bad!

julian-klode:
I think you want to use get_b or what it's called in the config object that returns a boolean so that it parses everything correctly.

anarcat:
> I think you want to use get_b or what it's called in the config object that returns a boolean so that it parses everything correctly.

It's a boolean? That wasn't quite clear to me, reading the documentation... I thought it could be a time delta (e.g. `1d`) as well, which is why I compare against `"0"`, as a string.

julian-klode:
Yes indeed, find_b sorry. Wasn't on my laptop so didn't look it up :)

anarcat:
thanks, fixed.

SuperQ (approved):

PR #185, opened 2023-10-21
adding chrony.sh to parse chrony NTP clients

allamiro:
Chrony.sh script in bash to parse chrony NTP clients data  since some environment cant have python installed on some servers.


SuperQ:
Why not https://github.com/SuperQ/chrony_exporter?

PR #195, opened 2023-12-15
Add needrestart_info.py to monitore needrestart.

RomainMou:
Hi,

I've made a small script to retrieve some metrics from `needrestart`, and perhaps it could be useful for others?

The script runs `needrestart -b` (batch mode), parses the output, and displays some metrics. It looks like this:
```
# HELP needrestart_timestamp information about the version and when it was last run
# TYPE needrestart_timestamp gauge
needrestart_timestamp{version="3.5"} 1.702655157e+09
# HELP needrestart_kernel_status information about the kernel status
# TYPE needrestart_kernel_status gauge
needrestart_kernel_status{current="5.10.0-26-amd64",expected="5.10.0-26-amd64"} 1.0
# HELP needrestart_services_count number of services requiring a restart
# TYPE needrestart_services_count gauge
needrestart_services_count 0.0
# HELP needrestart_containers_count number of containers requiring a restart
# TYPE needrestart_containers_count gauge
needrestart_containers_count 0.0
# HELP needrestart_sessions_count number of sessions requiring a restart
# TYPE needrestart_sessions_count gauge
needrestart_sessions_count 0.0
```

### DIFF
--- a/needrestart_info.py
+++ b/needrestart_info.py
@@ -145,10 +145,17 @@ def write_sessions(registry, needrestart_data):
 def main():
     registry = CollectorRegistry()
 
-    needrestart_output = subprocess.run(
-        ["needrestart", "-b"], capture_output=True, text=True
-    ).stdout
-    needrestart_data = NeedRestartData(needrestart_output)
+    try:
+        needrestart_output = subprocess.run(
+            ["needrestart", "-b"], capture_output=True, text=True, check=True
+        ).stdout
+        needrestart_data = NeedRestartData(needrestart_output)
+    except subprocess.CalledProcessError as e:
+        print(f"Error executing needrestart:\n{e}")
+        sys.exit(1)
+    except Exception as e:
+        print(f"An unexpected error occurred:\n{e}")
+        sys.exit(1)
 
     write_timestamp(registry, needrestart_data)
     write_kernel(registry, needrestart_data)
--- a/needrestart_info.py
+++ b/needrestart_info.py
@@ -33,7 +33,7 @@ class KernelStatus(Enum):
 class MicroCodeStatus(Enum):
     UNKNOWN = 0
     CURRENT = 1
-    OBSELETE = 2
+    OBSOLETE = 2
 
 
 class NeedRestartData:
--- a/needrestart_info.py
+++ b/needrestart_info.py
@@ -11,6 +11,7 @@ Dependencies: python >= 3.5, python3-prometheus-client, needrestart
 Authors: RomainMou
 """
 
+import sys
 import time
 import subprocess
 from collections import Counter
--- a/needrestart_info.py
+++ b/needrestart_info.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+#
+#
+# Description: Expose metrics from needrestart.
+#
+# This script runs needrestart in batch mode. It will never ask for input
+# and will never restart or upgrade anything.
+#
+# Dependencies: python >= 3.5, python3-prometheus-client, needrestart
+#
+# Authors: RomainMou
+
+import time
+import subprocess
+from collections import Counter
+from enum import Enum
+from prometheus_client import (
+    CollectorRegistry,
+    Gauge,
+    generate_latest,
+)
+
+
+class KernelStatus(Enum):
+    UNKNOWN = 0
+    CURRENT = 1
+    ABI_UPGRADE = 2
+    VERSION_UPGRADE = 3
+
+
+class MicroCodeStatus(Enum):
+    UNKNOWN = 0
+    CURRENT = 1
+    OBSELETE = 2
+
+
+class NeedrestartParser:
+    def __init__(self, needrestart_output):
+        # Some default value
+        self.timestamp = int(time.time())
+        self.version = None
+        self.kernel_status = None
+        self.microcode_status = None
+        self.kernel_current_version = ""
+        self.kernel_expected_version = ""
+        self.microcode_current_version = ""
+        self.microcode_expected_version = ""
+        needrestart_counter = Counter()
+
+        # Parse the cmd output
+        for line in needrestart_output.stdout.decode().splitlines():
+            key, value = line.split(": ")
+            if key == "NEEDRESTART-VER":
+                self.version = value
+            # Kernel informations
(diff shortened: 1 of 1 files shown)


needrestart_info.py
```diff
@@ -0,0 +1,160 @@
...
+def _write_sessions(registry, needrestart_data):
+    g = Gauge(
+        "needrestart_sessions_count",
+        "number of sessions requiring a restart",
+        registry=registry,
+    )
+    g.set(needrestart_data.sessions_count)
+
+
+def _main():
```
funkyfuture, line 143: imo, this doesn't need to be marked as private. per convention. this modules isn't intended to be used as library, so i'd question any marking of functions as private.

needrestart_info.py
```diff
@@ -0,0 +1,160 @@
...
+    registry = CollectorRegistry()
+    needrestart_data = NeedrestartParser(
+        subprocess.run(["needrestart", "-b"], stdout=subprocess.PIPE)
```
funkyfuture, line 146:
```suggestion
    needrestart_output = subprocess.run(["needrestart", "-b"], capture_output=True, text=True).stdout
    needrestart_data = NeedrestartParser(needrestart_output)
```

i think that makes it clearer that the process' output is to be used, the `text` option makes is unnecessary to decode it in the parsing logic.

the extra statement for capturing the output makes it better comprehensible imo.

needrestart_info.py
```diff
@@ -0,0 +1,160 @@
...
+#
+# Dependencies: python >= 3.5, python3-prometheus-client, needrestart
+#
+# Authors: RomainMou
+
+import time
+import subprocess
+from collections import Counter
+from enum import Enum
+from prometheus_client import (
```
funkyfuture, line 17:
```suggestion

from prometheus_client import (
```

per convention, 3rd party module imports are separated w/ an empty line from the ones from stdlib.

needrestart_info.py
```diff
@@ -0,0 +1,160 @@
...
+#
+#
+# Description: Expose metrics from needrestart.
+#
+# This script runs needrestart in batch mode. It will never ask for input
+# and will never restart or upgrade anything.
+#
+# Dependencies: python >= 3.5, python3-prometheus-client, needrestart
+#
+# Authors: RomainMou
```
funkyfuture, line 11:
```suggestion

"""
Description: Expose metrics from needrestart.

This script runs needrestart in batch mode. It will never ask for input
and will never restart or upgrade anything.

Dependencies: python >= 3.5, python3-prometheus-client, needrestart

Authors: RomainMou
"""
```

This is considered as module docstring by Python interpreters.

RomainMou:
Hello @funkyfuture, thanks for all your feedback, they are very appreciate for a beginner like me.
I've try to take all of them into account.

funkyfuture (approved):
lgtm, i've been using it for some weeks now.

RomainMou:
Is there anything missing to merge this PR? Anything I can do?

needrestart_info.py
```diff
@@ -0,0 +1,164 @@
...
+    UNKNOWN = 0
+    CURRENT = 1
+    ABI_UPGRADE = 2
+    VERSION_UPGRADE = 3
+
+
+class MicroCodeStatus(Enum):
+    UNKNOWN = 0
+    CURRENT = 1
+    OBSELETE = 2
```
dswarbrick, line 36: Spelling
```suggestion
    OBSOLETE = 2
```

needrestart_info.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+    g.set(needrestart_data.sessions_count)
+
+
+def main():
+    registry = CollectorRegistry()
+
+    needrestart_output = subprocess.run(
+        ["needrestart", "-b"], capture_output=True, text=True
+    ).stdout
```
dswarbrick, line 150: There ought to be some error handling here. What if the command fails to run, i.e. returns non-zero exit status, or what if `subprocess.run()` throws an exception?

dswarbrick (changes requested):

dswarbrick:
Please also check that your metric names conform to the Prometheus best practice guide: https://prometheus.io/docs/practices/naming/

dswarbrick (approved):

calestyo:
Just for the records: There's has also been something coded in `needrestart` itself: https://github.com/liske/needrestart/issues/291

anarcat:
> Just for the records: There's has also been something coded in `needrestart` itself: [liske/needrestart#291](https://github.com/liske/needrestart/issues/291)

should we consider removing this code, considering there's now something upstream? as one of those who worked on the upstream code, i would much rather see people collaborate there instead of duplicating the efforts here...

lelutin:
> > Just for the records: There's has also been something coded in `needrestart` itself: [liske/needrestart#291](https://github.com/liske/needrestart/issues/291)
> 
> should we consider removing this code, considering there's now something upstream? as one of those who worked on the upstream code, i would much rather see people collaborate there instead of duplicating the efforts here...

fwiw the new `needrestart -o` has been merged upstream, but it'll take a while for that version to be included in a new distro release anywhere. for example in debian it's going to be available in Trixie when it's released.

So I'm guessing this script will still be useful for some folks until the new option becomes widely available. Maybe we should mark this script as deprecated somehow and wait for the new needrestart option to become more widely available before actually removing it

anarcat:
On 2025-01-09 07:34:34, Gabriel Filion wrote:
>> > Just for the records: There's has also been something coded in `needrestart` itself: [liske/needrestart#291](https://github.com/liske/needrestart/issues/291)
>> 
>> should we consider removing this code, considering there's now something upstream? as one of those who worked on the upstream code, i would much rather see people collaborate there instead of duplicating the efforts here...
>
> fwiw the new `needrestart -o` has been merged upstream, but it'll take a while for that version to be included in a new distro release anywhere. for example in debian it's going to be available in Trixie when it's released.
>
> So I'm guessing this script will still be useful for some folks until the new option becomes widely available. Maybe we should mark this script as deprecated somehow and wait for the new needrestart option to become more widely available before actually removing it

Yeah, that makes sense. In that case Someone should at least make sure
the metrics are either similar enough that they're compatible, or that
the namespaces are distinct to provide a clean transition. :)

PR #201, opened 2024-01-29
storcli.py: don't error out on unsupported features

jcpunk:
Tested hardware: `LSI SAS3008 PCI-Express Fusion-MPT SAS-3 (rev 02)`

This integrated RAID chip doesn't emit stats for these features.

### DIFF
--- a/storcli.py
+++ b/storcli.py
@@ -340,27 +340,33 @@ def create_metrics_of_physical_drive(physical_drive, detailed_info_array, contro
         attributes = info[drive_identifier + " Device attributes"]
         settings = info[drive_identifier + " Policies/Settings"]
 
-        metrics["pd_shield_counter"].labels(controller_index, enclosure, slot).set(
-            state["Shield Counter"]
-        )
-        metrics["pd_media_errors"].labels(controller_index, enclosure, slot).set(
-            state["Media Error Count"]
-        )
-        metrics["pd_other_errors"].labels(controller_index, enclosure, slot).set(
-            state["Other Error Count"]
-        )
-        metrics["pd_predictive_errors"].labels(controller_index, enclosure, slot).set(
-            state["Predictive Failure Count"]
-        )
+        if state["Shield Counter"] != "N/A":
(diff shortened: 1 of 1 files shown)


LuminatiHD:
While you're at it, could you do a similar thing for the drive temperature? We had a similar problem there with our drive.

jcpunk:
I'm not sure which elements you need tweaked...

LuminatiHD:
`state["Drive Temperature"]` returns "N/A" for some of our drives (Line 393). I have a merge request up on this as well, but it might be neater to have everything part of the same Pull request.

jcpunk:
@dswarbrick Could you review this PR (and others for storcli.py)?

dswarbrick:
@jcpunk Are these "N/A" and "Unknown" values being emitted by storcli for actual online and spun-up drives? This seems like a pretty big oversight by LSI to omit these data, considering that most of them are mandatory SMART items. Are you sure it's not just a case of a spun-down / offline / power-save / standby disk?

storcli.py
```diff
@@ -340,27 +340,34 @@ def create_metrics_of_physical_drive(physical_drive, detailed_info_array, contro
...
-        )
+        if attributes["Link Speed"] != "Unknown":
+            metrics["pd_link_speed"].labels(controller_index, enclosure, slot).set(
+                attributes["Link Speed"].split(".")[0]
+            )
+        if attributes["Device Speed"] != "Unknown":
+            metrics["pd_device_speed"].labels(controller_index, enclosure, slot).set(
+                attributes["Device Speed"].split(".")[0]
+            )
+
```
dswarbrick, line 370: nit: surplus blank line

jcpunk: Fixed.

dswarbrick:
Since storcli is closed source, we don't have any way to dig deeper into why a spun-up drive is not reporting basic info like this. I'm okay with merging this, but we might want to consider a helper function in future to reduce some of the if-statement boilerplate code.

PR #206, opened 2024-02-17
smartmon.sh: Support sntrealtek device type

kjetijor:
sntrealtek is a NVMe device behind a Realtek RTL9210 USB to NVMe, output Looks like regular NVMe devices.


dswarbrick:
The smartmon.sh script is essentially deprecated and superseded by smartmon.py. Since the smartmon.sh script will be removed at some point in future, do you wish to also add equivalent functionality to smartmon.py?

kjetijor:
Having looking briefly at smartmon.py, we wouldn't need to change that, as it uses different heuristics for deciding whether or not to probe a device. If smartmon.sh is "more or less" deprecated, we can probably just close this, and I'll move on to smartmon.py instead of patching on smartmon.sh.

PR #210, opened 2024-04-18
btrfs_stats.py: Precompiled regular expressions improve performance

manas-rust:
**Performance Improvement:** The regular expression `DEVICE_PATTERN` is precompiled and stored as a function-level variable, ensuring it doesn't need to be recompiled for each iteration in the loop

### DIFF
--- a/btrfs_stats.py
+++ b/btrfs_stats.py
@@ -14,6 +14,7 @@ from prometheus_client import CollectorRegistry, Gauge, generate_latest
 
 DEVICE_PATTERN = re.compile(r"^\[([^\]]+)\]\.(\S+)\s+(\d+)$", re.MULTILINE)
 
+
 def get_btrfs_mount_points():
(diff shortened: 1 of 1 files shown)


btrfs_stats.py
```diff
@@ -12,6 +12,9 @@
 from prometheus_client import CollectorRegistry, Gauge, generate_latest
 
 
+DEVICE_PATTERN = re.compile(r"^\[([^\]]+)\]\.(\S+)\s+(\d+)$", re.MULTILINE)
```
dswarbrick, line 15: Is there some reason why you make this `re.MULTILINE`? From what I can see, it's unnecessary, as the `get_btrfs_errors` function parses the output one line at a time.

manas-rust: Thank you for bringing this to my attention.

dswarbrick (approved):

PR #212, opened 2024-05-08
lint: format yum.sh

manas-rust:
format yum.sh


dswarbrick:
Can you be more specific about your chosen method of formatting? I'm assuming that you used `shfmt --indent 2`.

manas-rust:
> Can you be more specific about your chosen method of formatting? I'm assuming that you used `shfmt --indent 2`.

```bash
shfmt -i 2 -l -w yum.sh
```

PR #217, opened 2024-07-10
Add helium_level to the list of smartmon attributes to report.

benlivengood:
Some of my HGST/WD hard drives export `22 Helium_Level` as an attribute, and Toshiba drives use `23 Helium_Condition_Lower` and `24 Helium_Condition_Upper` and I want to monitor them.

I've tested this change locally to make sure the new metrics appear in prometheus.


benlivengood:
I've since obtained some new drives and found this old PR sitting here!  Updated to include smartmon.py

Issue #226, opened 2024-11-17
nvme-cli v2.11 breaks compatibility with nvme textfile collector


dswarbrick:
nvme-cli v2.11 now defaults to outputting _verbose_ JSON output, with apparently no way to revert back to the previous "terse" output.

For the command `nvme list -o json`, the previous output (with nvme-cli v2.10) resembled the following:
```json
{
  "Devices":[
    {
      "NameSpace":1,
      "DevicePath":"/dev/nvme0n1",
      "GenericPath":"/dev/ng0n1",
      "Firmware":"3B4QFXO7",
      "ModelNumber":"Samsung SSD 980 500GB",
      "SerialNumber":"S64D...",
      "UsedBytes":46430703616,
      "MaximumLBA":976773168,
      "PhysicalSize":500107862016,
      "SectorSize":512
    }
  ]
}
```
With nvme-cli v2.11, that exact same command now results in this:
```json
{
  "Devices":[
    {
      "HostNQN":"nqn.2014-08.org.nvmexpress:uuid:235adbb6-...",
      "HostID":"12718199-...",
      "Subsystems":[
        {
          "Subsystem":"nvme-subsys0",
          "SubsystemNQN":"nqn.1994-11.com.samsung:nvme:980M.2:S64D...",
          "Controllers":[
            {
              "Controller":"nvme0",
              "Cntlid":"5",
              "SerialNumber":"S64D...",
              "ModelNumber":"Samsung SSD 980 500GB",
              "Firmware":"3B4QFXO7",
              "Transport":"pcie",
              "Address":"0000:02:00.0",
              "Slot":"",
              "Namespaces":[
                {
                  "NameSpace":"nvme0n1",
                  "Generic":"ng0n1",
                  "NSID":1,
                  "UsedBytes":46439002112,
                  "MaximumLBA":976773168,
                  "PhysicalSize":500107862016,
                  "SectorSize":512
                }
              ],
              "Paths":[]
            }
          ],
          "Namespaces":[]
        }
      ]
    }
  ]
}
```
This obviously breaks the nvme textfile collector, as it is not expecting to parse the verbose output.

As far as I can tell, this was annotated in the [v2.11 release notes](https://github.com/linux-nvme/nvme-cli/releases/tag/v2.11) as:
>       nvme-print-json: display only verbose output

[...]

PR #227, opened 2024-11-19
nvme_metrics: support nvme-cli v2.11+ verbose JSON output

dswarbrick:
nvme-cli v2.11 introduced a breaking change by forcing JSON output to always be verbose, resulting in a significantly different output structure.

Also drop the obsolete nvme_metrics.sh shell script collector, which will not work with nvme-cli v2.11+ (unless it were also refactored).

Fixes: #226


dswarbrick:
This PR aims solely to support nvme-cli v2.11+, which produces more verbose and deeply nested JSON output than previous versions of the tool.

Some decisions that were made in the past regarding the nvme_metrics collector (i.e., the original shell script version) do not translate so well to this deeply nested information, which will ideally require some breaking changes to the metrics produced, in terms of metric name and labels.

At the same time, there is a growing likelihood that this collector may be run in environments which have non-local NVMe devices, e.g. NVMe over fabrics or NVMe over TCP, which should perhaps be distinguished with a `transport` label in their info metric.

SuperQ (approved):
Thanks!

PR #228, opened 2024-11-19
nvme_metrics: refactor metrics to better fit the verbose / nested JSON output of nvme-cli

dswarbrick:


dswarbrick:
@sbates130272 Can you perhaps confirm that the `nvme smart-log` command is indeed controller-specific, not namespace-specific? I noticed that nvme-cli will work with either e.g. `/dev/nvme0` or `/dev/nvme0n1`, but as far as I can tell, the data returned are controller-specific. I don't _think_ counters such as `data_units_written` are tracked per namespace.

I would like to refactor a few of the metrics that we currently expose as "per-device" (which is a misnomer), and instead expose them per controller. This will unfortunately affect about 15 metrics, but I feel that it is important to get this right.

For example, this:
```
nvme_power_on_hours_total{device="nvme0n1"} 16170.0
```
would become:
```
nvme_power_on_hours_total{controller="nvme0"} 16170.0
```
... as I don't think it's possible for a _namespace_ to be powered on for a different number of hours than another namespace on the same controller.

The metric / labels implemented by this collector hark back to the original nvme_metrics.sh collector, and were based on the wrong assumption that all NVMe devices are single-namespace - probably due to only having consumer-grade SSDs to test it on at the time.

sbates130272:
@dswarbrick I did a quick check and depending on the NVMe device it may support SMART log pages on a per namespace basis or it may not. The ability to do this on a per namespace basis is advertised via a field in the controller. Here is the help information for the ```nvme smart-log``` command.
```
NVME-SMART-LOG(1)                                                              NVMe Manual                                                             NVME-SMART-LOG(1)

NAME
       nvme-smart-log - Send NVMe SMART log page request, returns result and log

SYNOPSIS
       nvme smart-log <device> [--namespace-id=<nsid> | -n <nsid>]
                               [--raw-binary | -b]
                               [--output-format=<fmt> | -o <fmt>] [--verbose | -v]

DESCRIPTION
       Retrieves the NVMe SMART log page from an NVMe device and provides the returned structure.

       The <device> parameter is mandatory and may be either the NVMe character device (ex: /dev/nvme0), or a namespace block device (ex: /dev/nvme0n1).

       On success, the returned smart log structure may be returned in one of several ways depending on the option flags; the structure may parsed by the program and
       printed in a readable format or the raw buffer may be printed to stdout for another program to parse.

OPTIONS
       -n <nsid>, --namespace-id=<nsid>
           Retrieve the SMART log for the given nsid. This is optional and its success may depend on the device’s capabilities to provide this log on a per-namespace
           basis (see the NVMe Identify Controller for this capability). The default nsid to use is 0xffffffff for the device global SMART log.

       -b, --raw-binary
           Print the raw SMART log buffer to stdout.

       -o <fmt>, --output-format=<fmt>
           Set the reporting format to normal, json or binary. Only one output format can be used at a time.

       -v, --verbose
           Increase the information detail in the output.

EXAMPLES
       •   Print the SMART log page in a human readable format:

               # nvme smart-log /dev/nvme0

       •   Print the raw SMART log to a file:
```

sbates130272:
@dswarbrick is it worth sending an email to the [linux-nvme][ref-nvme-mail] mailing list so that the people on that list can chime in on this?

[ref-nvme-mail]: https://lists.infradead.org/pipermail/linux-nvme/

sbates130272:
@dswarbrick I took a look at the latest version of the [NVMe specification][ref-nvme-spec] and the [relevant code in nvme-cli][ref-nvme-cli] and it looks like nvme-cli will return a controller-based SMART page unless the  ```--namespace-id``` is set. In which case it will return a namespace-based SMART log (or an error if the device does not support namespace-based SMART log pages).

[ref-nvme-spec]: https://nvmexpress.org/specifications/
[ref-nvme-cli]: https://github.com/linux-nvme/nvme-cli/blob/master/nvme.c#L519

dswarbrick:
@sbates130272 Thanks for researching that. So I think that, as I alluded to with the FIXME in the current code, it would make sense to just fetch the global SMART log once per controller, i.e., without specifying a namespace, and change the `device` label to `controller`. If there is a strong need in future, we could fetch it per namespace, with the addition of a `nsid` label in the metric. The key thing is clear up the misunderstanding with the e.g. `device="nvme0n1"` labels.

dswarbrick:
@sbates130272 How does this PR look to you now?

sbates130272 (approved):

nvme_metrics.py
```diff
@@ -24,102 +24,109 @@
 
 metrics = {
     # fmt: off
+    "nvmecli": Info(
```
sbates130272, line 27: NIT.  Add a comment to make it clear this section is for host-specific metrics
```
# Host-specific metrics
```

nvme_metrics.py
```diff
@@ -159,75 +166,73 @@ def main():
...
+                        ctrl_dev,
                     )
 
-                    # NVMe reports temperature in kelvins; convert it to degrees Celsius.
-                    metrics["temperature"].labels(device_name).set(smart_log["temperature"] - 273)
+                    metrics["sector_size"].labels(ns_dev).set(ns["SectorSize"])
+                    metrics["physical_size"].labels(ns_dev).set(ns["PhysicalSize"])
+                    metrics["used_bytes"].labels(ns_dev).set(ns["UsedBytes"])
+
+                # Most SSDs (perhaps _all_ consumer grade SSDs) only contain a single namespace.
```
sbates130272, line 199: This comment may not be accurate. A vast majority of NVMe SSDs live in data-centers and those may have multiple namespaces. It may make sense to drop the first sentence.

nvme_metrics.py
```diff
@@ -159,75 +166,73 @@ def main():
...
+                # Fetch the device global SMART log by omitting any --namespace-id flag.
```
sbates130272, line 200: s/device/controller/ and s/any/the/

sbates130272 (changes requested):
Just a few nits. Thanks for doing this!

PR #231, opened 2025-01-09
add helper script to extract timestamps from files

anarcat:
This is a pretty simple tool, but we've found this useful to check some diagnostics. We've also found alternatives like:

https://github.com/Puppet-Finland/prometheus-file-age-metrics/blob/main/create-file-age-metrics.sh

... which gives you a metric like "out of date files", which didn't like, because the threshold lives in the script instead of the alerting layer (which we prefer).

The closest thing to this is the filestat exporter:

https://github.com/michael-doubez/filestat_exporter

... which provides way more metrics and seems overengineered for our purpose.


anarcat:
this is a draft because we haven't tested the code yet. @lelutin will deploy this and we'll report back!

path-timestamp.sh
```diff
@@ -0,0 +1,16 @@
...
+# Usage: add this to crontab:
+#
+# */5 * * * * prometheus path-timestamp.sh /var/lib/prometheus | sponge /var/lib/node_exporter/path-timestamp.prom
+#
+# Author: Antoine Beaupré <anarcat@debian.org>
+
+echo "# HELP node_path_modification_timestamp_seconds Last change timestamp"
+echo "# TYPE node_path_modification_timestamp_seconds gauge"
+for path in "$@"; do
+    printf 'node_path_modification_timestamp_seconds{path="%s"}' "$path"
```
lelutin, line 14:
```suggestion
    printf 'node_path_modification_timestamp_seconds{path="%s"} ' "$path"
```

there's currently no space between the list of labels and the value

lelutin (changes requested):

path-timestamp.sh
```diff
@@ -0,0 +1,16 @@
...
+    stat -c %Y "$path"
```
lelutin, line 15:
```suggestion
    if ! stat -c %Y "$path"; then
      # clear line for next metric
      echo 0
    fi
```

If a file does not exist, this line can cause the output to become invalid like the following:

~~~
$ ./path-timestamp.sh /home/tor/dev/tpa/tor-puppet/hiera.yaml environment.conf path-timestamp.sh ext_fs_error_count.sh 
# HELP node_path_modification_timestamp_seconds Last change timestamp
# TYPE node_path_modification_timestamp_seconds gauge
node_path_modification_timestamp_seconds{path="/home/tor/dev/tpa/tor-puppet/hiera.yaml"} 1734046134
node_path_modification_timestamp_seconds{path="environment.conf"} 1717708435
node_path_modification_timestamp_seconds{path="path-timestamp.sh"} 1736455719
node_path_modification_timestamp_seconds{path="ext_fs_error_count.sh"} 1736452903
~~~

or without stderr we have:

~~~
$ ./path-timestamp.sh /home/tor/hiera.yaml environment.conf path-timestamp.sh ext_fs_error_count.sh 2>/dev/null
# HELP node_path_modification_timestamp_seconds Last change timestamp
# TYPE node_path_modification_timestamp_seconds gauge
node_path_modification_timestamp_seconds{path="/home/tor/hiera.yaml"} node_path_modification_timestamp_seconds{path="environment.conf"} 1717708435
node_path_modification_timestamp_seconds{path="path-timestamp.sh"} 1736455719
node_path_modification_timestamp_seconds{path="ext_fs_error_count.sh"} 1736452903

lelutin (changes requested):

path-timestamp.sh
anarcat: fixed

path-timestamp.sh
anarcat: i implemented a different solution for this, where we set the timestamp to zero if stat(1) fails, which will allow us to alert on those cases as well.

lelutin: great, so we had the same idea wrt setting it to 0, but your solution is more compact. and it's good that the comment above mentions why the value can be set to 0. :100:

lelutin (approved):
LGTM!

path-timestamp.sh
anarcat: oh. oops. sorry, i misread your suggested change and interpreted it as:

```
exit 0
```

not:

```
echo 0
```

as you wrote.

it seems like we're in violent agreement then! :)

what do you think of silencing the error from the stat failure? should we let those through? i was thinking maybe it would make diagnostics easier in case of failures, but then it might mess up the output if, for some reason stupid reason i can't take of, (e.g.) sponge(1) takes in stderr alongside stdout...

Issue #220, opened 2024-08-21
export more information from apt_info.py: package count, obsolete, broken packages


anarcat:
Hi,

We're migrating away from icinga here and we have an in-house check that provides us with quite useful metrics.

`apt_info.py` has this:

```
# /usr/share/prometheus-node-exporter-collectors/apt_info.py
# HELP apt_upgrades_pending Apt packages pending updates by origin.
# TYPE apt_upgrades_pending gauge
apt_upgrades_pending{origin="",arch=""} 0
# HELP apt_upgrades_held Apt packages pending updates but held back.
# TYPE apt_upgrades_held gauge
apt_upgrades_held{origin="",arch=""} 0
# HELP apt_autoremove_pending Apt packages pending autoremoval.
# TYPE apt_autoremove_pending gauge
apt_autoremove_pending 21
# HELP node_reboot_required Node reboot is required for software updates.
# TYPE node_reboot_required gauge
node_reboot_required 0
```

our `dsa-check-packages` has this:

```
# dsa-check-packages 
OK: 618 ok, 4 rc
618 packages current.
4 packages removed but not purged: linux-image-6.1.0-18-amd64, linux-image-6.1.0-17-amd64, linux-image-6.1.0-21-amd64, linux-image-6.1.0-20-amd64
|obs_loc=0;1;5;0 outdated=0;1;5;0 current=618;;;0 obs_ign=0;;;0 rm_unprg=4;;;0 hold=0;;;0 prg_conf=0;1;;0
```

that is:

- [x] held packages (`apt_upgrades_held`)
- [ ] obsolete packages, including a way to exclude packages (not covered!)
- [x] outdated packages (`apt_upgrades_pending`)
- [ ] current packages (not covered!)
- [ ] removed packages with leftover configuration ("removed but not purged", not covered)
- [ ] purged packages with leftove configuration (!?) (not covered)

the "obsolete packages" is the most important, as those are packages that might not be covered by security upgrade and a manual configuration...

i plan on adding those into apt_info.py through a merge request unless someone objects in a coming weeks. please let me know.

anarcat:
not part of my use case, but sure, that would be nice.

how do you tell those apart? how does apt know those shouldn't be updated?

in this case, for example, could you show the output of `apt-cache policy cloud-init`?

dswarbrick:
[...]

PR #234, opened 2025-01-22
collect more apt metrics (Closes: #220)

lelutin:
This patch series adds some more information about:

* total installed packages, per arch and origin
* obsolete packages -- those without a candidate, an origin or a with only one origin that points to local storage

It also adds filtering for packages that are marked for phasing upgrades. In this situation, the upgrades are not really pending anymore but just delayed until they are actually processed.

### DIFF
--- a/apt_info.py
+++ b/apt_info.py
@@ -35,8 +35,13 @@ def _convert_candidates_to_upgrade_infos(candidates):
     changes_dict = collections.defaultdict(lambda: collections.defaultdict(int))
 
     for candidate in candidates:
+        # The 'now' archive only shows that packages are not installed. We tend
+        # to filter the candidates on those kinds of conditions before reaching
+        # here so here we don't want to include this information in order to
+        # reduce noise in the data.
         origins = sorted(
-            {f"{o.origin}:{o.codename}/{o.archive}" for o in candidate.origins}
+            {f"{o.origin}:{o.codename}/{o.archive}" for o in candidate.origins
+             if o.archive != 'now'}
         )
         changes_dict[",".join(origins)][candidate.architecture] += 1
 
@@ -105,6 +110,17 @@ def _write_autoremove_pending(registry, cache):
     g.set(len(autoremovable_packages))
 
 
+def _write_installed_packages_per_origin(registry, cache):
+    installed_packages = {p.candidate for p in cache if p.is_installed}
+    per_origin = _convert_candidates_to_upgrade_infos(installed_packages)
+
+    if per_origin:
+        g = Gauge('apt_packages_per_origin_count', "Number of packages installed per origin.",
+                  ['origin', 'arch'], registry=registry)
+        for o in per_origin:
+            g.labels(o.labels['origin'], o.labels['arch']).set(o.count)
+
+
 def _write_cache_timestamps(registry):
     g = Gauge('apt_package_cache_timestamp_seconds', "Apt update last run time.", registry=registry)
     apt_pkg.init_config()
@@ -138,6 +154,7 @@ def _main():
     _write_held_upgrades(registry, cache)
     _write_obsolete_packages(registry, cache)
     _write_autoremove_pending(registry, cache)
+    _write_installed_packages_per_origin(registry, cache)
     _write_cache_timestamps(registry)
     _write_reboot_required(registry)
     print(generate_latest(registry).decode(), end='')
--- a/apt_info.py
+++ b/apt_info.py
@@ -84,6 +84,20 @@ def _write_held_upgrades(registry, cache):
             g.labels(change.labels['origin'], change.labels['arch']).set(change.count)
 
 
+def _write_obsolete_packages(registry, cache):
+    # This corresponds to the apt filter "?obsolete"
+    obsoletes = [p for p in cache if p.is_installed and (
+                  p.candidate is None or
+                  not p.candidate.origins or
+                  (len(p.candidate.origins) == 1 and
+                   p.candidate.origins[0].origin in ['', "/var/lib/dpkg/status"])
+                )]
+
+    g = Gauge('apt_packages_obsolete_count', "Apt packages which are obsolete",
+              registry=registry)
+    g.set(len(obsoletes))
+
+
 def _write_autoremove_pending(registry, cache):
     autoremovable_packages = {p for p in cache if p.is_auto_removable}
     g = Gauge('apt_autoremove_pending', "Apt packages pending autoremoval.",
@@ -122,6 +136,7 @@ def _main():
     registry = CollectorRegistry()
     _write_pending_upgrades(registry, cache)
     _write_held_upgrades(registry, cache)
+    _write_obsolete_packages(registry, cache)
     _write_autoremove_pending(registry, cache)
     _write_cache_timestamps(registry)
     _write_reboot_required(registry)
--- a/apt_info.py
+++ b/apt_info.py
@@ -25,7 +25,9 @@ Authors: Kyle Fazzari <kyrofa@ubuntu.com>
 import apt
 import apt_pkg
 import collections
+import logging
 import os
+import sys
 from prometheus_client import CollectorRegistry, Gauge, generate_latest
 
 _UpgradeInfo = collections.namedtuple("_UpgradeInfo", ["labels", "count"])
@@ -62,6 +64,12 @@ def _write_pending_upgrades(registry, cache):
     candidates = {
         p.candidate for p in cache if p.is_upgradable and not p.phasing_applied
     }
+    for candidate in candidates:
+        logging.debug(
+            "pending upgrade: %s / %s",
+            candidate.package,
+            candidate.architecture,
+        )
     upgrade_list = _convert_candidates_to_upgrade_infos(candidates)
 
     if upgrade_list:
@@ -80,6 +88,12 @@ def _write_held_upgrades(registry, cache):
             and not p.phasing_applied
         )
     }
+    for candidate in held_candidates:
+        logging.debug(
+            "held upgrade: %s / %s",
+            candidate.package,
+            candidate.architecture,
+        )
     upgrade_list = _convert_candidates_to_upgrade_infos(held_candidates)
 
     if upgrade_list:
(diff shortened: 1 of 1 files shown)


apt_info.py
```diff
@@ -35,8 +35,12 @@ def _convert_candidates_to_upgrade_infos(candidates):
     changes_dict = collections.defaultdict(lambda: collections.defaultdict(int))
 
     for candidate in candidates:
+        # The 'now' archive only shows that packages are not installed. We tend
+        # to filter the candidates on those kinds of conditions before reaching
+        # here so here we don't want to include this information in order to
+        # reduce noise in the data.
         origins = sorted(
-            {f"{o.origin}:{o.codename}/{o.archive}" for o in candidate.origins}
+            {f"{o.origin}:{o.codename}/{o.archive}" for o in candidate.origins if o.archive !='now'}
```
anarcat, line 43: linter is unhappy with this one:

```suggestion
            {f"{o.origin}:{o.codename}/{o.archive}" for o in candidate.origins if o.archive != 'now'}
```

apt_info.py
```diff
@@ -69,7 +73,7 @@ def _write_pending_upgrades(registry, cache):
 def _write_held_upgrades(registry, cache):
     held_candidates = {
         p.candidate for p in cache
-        if p.is_upgradable and p._pkg.selected_state == apt_pkg.SELSTATE_HOLD
+        if p.is_upgradable and p._pkg.selected_state == apt_pkg.SELSTATE_HOLD and not p.phasing_applied
```
anarcat, line 76: linter also yells here, not sure how to fix that, perhaps:

```suggestion
        if (
            p.is_upgradable 
            and p._pkg.selected_state == apt_pkg.SELSTATE_HOLD 
            and not p.phasing_applied
        )
```

apt_info.py
anarcat: also, is the count still correct with this change? I imagine it just removes the needless origin, but did you make sure the number of install packages metric total still correctly reflect the actual installed count?

apt_info.py
```diff
@@ -80,13 +84,38 @@ def _write_held_upgrades(registry, cache):
...
+def _write_obsolete_packages(registry, cache):
+    # This corresponds to the apt filter "?obsolete"
+    obsoletes = [p for p in cache if p.is_installed and (
+                  p.candidate is None or
+                  not p.candidate.origins or
+                  (len(p.candidate.origins) == 1 and
+                   p.candidate.origins[0].origin in ['', "/var/lib/dpkg/status"])
+                )]
+
+    g = Gauge('apt_obsolete_packages', "Apt packages which are obsolete",
```
anarcat, line 96: i wonder about that metric name. I think, on its own, it makes sense, but when compared to the other metrics generated here, less so. All other metrics here do not have a `_packages` suffix, so  I wonder if we shouldn't change the name here to be more coherent with the other metrics, which are:

```
apt_upgrades_pending
apt_upgrades_held
apt_autoremove_pending
apt_package_cache_timestamp_seconds
node_reboot_required
```

(It should be noted that most of the metrics here do not follow the proper naming convention, except maybe `apt_package_cache_timestamp_seconds`, but that's probably because *I* added it recently. Extra credits for the `node_reboot_required` metric, coming completely out of left field...)

Anyways. I think I'm convincing myself that the name you picked is actually not that bad, but if we're going to break the (bad) convention that's already there, let's make it follow a *proper* convention, and let's call it `apt_packages_obsolete_count`, or maaybe `apt_obsolete_count`.

apt_info.py
```diff
@@ -80,13 +84,38 @@ def _write_held_upgrades(registry, cache):
...
               registry=registry)
     g.set(len(autoremovable_packages))
 
 
+def _write_installed_packages_per_origin(registry, cache):
+    installed_packages = {p.candidate for p in cache if p.is_installed}
+    per_origin = _convert_candidates_to_upgrade_infos(installed_packages)
+
+    if per_origin:
+        g = Gauge('apt_installed_per_origin', "Numberof ppt packages installed per origin.",
```
anarcat, line 113: similar comment here: either make this coherent with the obsolete metric, or make the obsolete metric coherent with this. :)

i would call it `apt_packages_per_origin_count`... heck, you could even call it `apt_packages_count`, then we could add more labels (e.g. `state=installed`) as we go along converting the other metrics... ;) but maybe that would be too ambiguous.

also, the description has a couple of typos here, maybe you mean:

```suggestion
        g = Gauge('apt_installed_per_origin', "Number of packages installed per origin.",
```

anarcat (changes requested):
great start, thanks for working on this!

could you give an example output before/after the change? i think you have a before in: https://gitlab.torproject.org/tpo/tpa/prometheus-alerts/-/issues/19#note_3151581

or at least a sample of the new metrics on a live system, to give us an idea of what it looks like?

apt_info.py
lelutin: huh interesting, I just saw the failed flake8 test in CI. I'm using pywright locally because it was simpler to integrate into my editor + plugins but it's showing nothing for the problems that flake8 found. I guess that means it's time I made the effort of switching away from pywright.

It'll be fixed in my next push

apt_info.py
lelutin: ah yes you're right. I like `apt_packages_obsolete_count`, let's go for that one

apt_info.py
lelutin: idk about going with `apt_packages_count` since some packages will be double+ counted since they can match more than one state. personally because of the name I would kind of expect the total sum for `apt_packages_count` across all label variants to be the total installed packages, so if we have duplication it won't reflect that correctly.

let's go with `apt_packages_per_origin_count`

anarcat: okay, then the metrics name is definitely better..  i was actually assuming we did *not* have duplicates and could use that metric to extract a total package count, good to know that isn't the case.

anarcat (approved):

lelutin:
I always for get to include the signoff lines in the commits, sorry for the noise.

@anarcat I've added two more commits on the patch series to complete the requested features: we can now enable debug logs with an env var and we can also exclude packages by name with a command-line argument.

apt_info.py
```diff
@@ -147,13 +176,18 @@ def _write_reboot_required(registry):
 
 
 def _main():
+    logging.basicConfig(stream=sys.stderr)
+    logger = logging.getLogger(__name__)
+    if os.environ.get("DEBUG"):
+        logger.setLevel(logging.DEBUG)
+
```
anarcat, line 192: this is going to sound like nitpicking, but i would avoid passing a `logger` all over the place.

just skip setting a logger object and call `logging.debug` directly, i would say. if we were in a larger, multi-file project with various modules and third-party libs, i would do it differently, but in this case... why bother?

if we *do* want to redact (say) logs from python-apt, then i would do that explicitly.

otherwise, if you really want a global logger, then do that. maybe something like:

```python
if __name__ == "__main__":
    logging.basicConfig(stream=sys.stderr)
    logger = logging.getLogger(__name__)
    if os.environ.get("DEBUG"):
        logger.setLevel(logging.DEBUG)
    _main()
```

that way you don't need to pass logger around... i understand a global might not be the best practice, but that's effectively what we're doing here anyways by passing the logger all over the place...

apt_info.py
```diff
@@ -147,13 +176,18 @@ def _write_reboot_required(registry):
 
 
 def _main():
+    logging.basicConfig(stream=sys.stderr)
```
anarcat, line 188: i think stream=sys.stderr is pretty much the default, why do we need to do this here?

apt_info.py
```diff
@@ -60,9 +61,11 @@ def _convert_candidates_to_upgrade_infos(candidates):
     return changes_list
 
 
-def _write_pending_upgrades(registry, cache, logger):
+def _write_pending_upgrades(registry, cache, logger, cli_args):
```
anarcat, line 64: again maybe nitpicking, but i would pass the exclude flag directly here, not the whole args namespace:

```python
def _write_pending_upgrades(registry, cache, logger, exclude):
```

anarcat:
some nitpicking on the code, otherwise LGTM.

apt_info.py
lelutin: oh this is not nitpicking it's a good recommendation, it'll remove some useless complexity in the code. I'll push a change that uses the root logger instead

apt_info.py
lelutin: blah it is indeed. I'll remove the param

anarcat (approved):
LGTM, thanks!

lelutin:
so just for info, we've been using this modified script in production for a while now and we have an alert set on hosts with obsolete packages, and we've recently seen some hosts pop up for this alert after some upgrades to debian trixie. so the metrics are doing what they should be doing!

anarcat:
ping, can someone look into this?

anarcat:
hello!

so just so you know, we are now maintaining our own out-of-tree fork of this file. since then, i optimized the code (it was taking about 10 seconds to run on some machines, and i sped it up by about an order of magnitude.

unfortunately, those are kind of big changes, and considering smaller changes like this one are not being merged in a timely manner, i'm a little demotivated to try to merge more complicated stuff in. here's the script as it stands on our end right now: [apt_info.py](https://paste.anarc.at/publish/2026-04-20-Wtc-rkCM4iZvbtxIQR0Y3h8AEU_N220NQyDhHkGGtzA/apt_info.py). 

i'm happy to contribute back the half dozen commits that bring us there here, but this merge request is a first step towards that, and I'm not going to bother trying if we can't even get this simpler one in. :)

apt_info.py
```diff
@@ -80,13 +107,58 @@ def _write_held_upgrades(registry, cache):
...
+        if package.candidate is None:
+            logging.debug("obsolete package with no candidate: %s", package)
+        else:
+            logging.debug(
+                "obsolete package: %s / %s",
+                package,
+                package.candidate.architecture,
+            )
+
+    g = Gauge('apt_packages_obsolete_count', "Apt packages which are obsolete",
```
SuperQ:
```suggestion
    g = Gauge('apt_packages_obsolete', "Apt packages which are obsolete",
```

apt_info.py
```diff
@@ -80,13 +107,58 @@ def _write_held_upgrades(registry, cache):
...
               registry=registry)
     g.set(len(autoremovable_packages))
 
 
+def _write_installed_packages_per_origin(registry, cache):
+    installed_packages = {p.candidate for p in cache if p.is_installed}
+    per_origin = _convert_candidates_to_upgrade_infos(installed_packages)
+
+    if per_origin:
+        g = Gauge('apt_packages_per_origin_count', "Number of packages installed per origin.",
```
SuperQ:
```suggestion
        g = Gauge('apt_packages_per_origin', "Number of packages installed per origin.",
```

SuperQ (approved):

anarcat:
whoohoo! thanks! but also sorry: we'll send you the optimization work next. ;)

SuperQ:
Yes, sorry for neglecting this project.

anarcat:
don't be sorry, thanks for the cleanup!!

anarcat:
> whoohoo! thanks! but also sorry: we'll send you the optimization work next. ;)

this is now available in https://github.com/prometheus-community/node-exporter-textfile-collector-scripts/pull/269

PR #237, opened 2025-04-11
Always write value for apt_upgrades_pending and apt_upgrades_held

mpkossen:
Always write value for apt_upgrades_pending and apt_upgrades_held to ensure monitoring works.

If there are no held packages or no pending upgrades, nothing is written to apt.prom, resulting in monitoring dashboards not picking up on these values. This patch fixes that by ensuring there's always a value for apt_upgrades_pending and apt_upgrades_held .


varac:
@mpkossen Can you please fix the GH actions Lint/flake8 error ? Unfortunatly the log is expired, so pushing again might trigger a new one.

varac:
@dswarbrick  Awesome, thanks for the quick response !

PR #239, opened 2025-05-13
[directory-size.sh] Support spaces in paths

aroberts:
The current code emits directory label only up to the first space, due to the default behavior of `awk`. This change pulls in the rest of the line output from `du` into the label, so directories containing spaces are properly captured

### DIFF
--- a/directory-size.sh
+++ b/directory-size.sh
@@ -6,8 +6,6 @@
 #
 # */5 * * * * prometheus directory-size.sh /var/lib/prometheus | sponge /var/lib/node_exporter/directory_size.prom
 #
-# awk logic from by https://stackoverflow.com/a/10221507
-#
 # Author: Antoine Beaupré <anarcat@debian.org>
 
 echo "# HELP node_directory_size_bytes Disk space used by some directories"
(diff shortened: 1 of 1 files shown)


dswarbrick:
I was about to suggest using `sed` instead, e.g.
```
sed -E 's/^([0-9]+)[[:blank:]](.+)$/node_directory_size_bytes{directory="\2"} \1/'
```
- however we also should handle directories with double-quotes in their names. Currently, these would cause malformed metrics to be printed, since the double-quotes need to be escaped.

aroberts:
@dswarbrick fair enough, implemented both

directory-size.sh
```diff
@@ -6,9 +6,14 @@
 #
 # */5 * * * * prometheus directory-size.sh /var/lib/prometheus | sponge /var/lib/node_exporter/directory_size.prom
 #
+# awk logic from by https://stackoverflow.com/a/10221507
+#
```
dswarbrick: These are obsolete.

dswarbrick:
Please sign off your commits to satisfy the DCO.

aroberts:
removed comment; rebased with signoffs

PR #240, opened 2025-05-17
smartmon.sh: locate smartctl tool on non-Linux systems

cristian-caloghera:
Allow execution of `smartctl` even if it is installed in an unexpected (for Linux) path.

Some OS, (e.g.: FreeBSD) do not install `smartctl` under the `usr/sbin/` path, but instead use another path like `/usr/local/sbin/`.
This commit makes use of the `which` command to locate the `smartctl` in the `PATH` and execute from there. As such it does not make any assumption where `smartctl` is installed, improving portability.

### DIFF
--- a/smartmon.sh
+++ b/smartmon.sh
@@ -167,7 +167,8 @@ format_output() {
     awk -F'{' "${output_format_awk}"
 }
 
-smartctl_version="$(/usr/sbin/smartctl -V | awk 'NR==1 && $1 == "smartctl" {print $2}')"
+smartctl_cmd=$(which smartctl)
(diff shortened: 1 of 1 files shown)


dswarbrick:
According to the manpage for `which`, it works "by searching the PATH for executable files matching the names of the arguments."

In other words, this could be accomplished by omitting the `which` stuff altogether, and simply using a non-qualified `smartctl` command, i.e. relying on the inherent functionality of the shell's PATH to find the command.

cristian-caloghera:
You are correct. No idea why it did not occur to me initially, but I have changed the code to call `smartctl` directly, i.e. without an absolute path.

PR #246, opened 2025-07-18
Add nvme-cli 2.3 compat

deajan:
Debian 12 (proxmox) ships with nvme-cli 2.3 which has no `--verbose` parameter for `smart-log` command.
We check if `--verbose` exists, and deactivate it if not present.

Patch has been tested on Debian 12 and RHEL 9 (with nvme-cli 2.11) in order to ensure there is no regression.

### DIFF
--- a/nvme_metrics.py
+++ b/nvme_metrics.py
@@ -131,6 +131,21 @@ metrics = {
 }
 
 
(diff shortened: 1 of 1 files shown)


deajan:
Closed in favor of #255

PR #247, opened 2025-07-18
Add --compat-disk-labels option

deajan:
So I use `smartmon.sh` for years now.
Migrating to `smartmon.py`, I really got frustrated because metrics like:
```
smartmon_device_info{disk="/dev/sda" [...]}
```
became
```
smartmon_device_info{device="/dev/sda", disk="0" [...]}
```
So basically I have to batch update all my instances, or develop new dashboards (I can't get disk or device variable in grafana, one has to choose).

So my proposition is to add `--compat-disk-labels` to `smartmon.py` in order for the latter to generate metrics with the same disk name as previously found in `smarmon.sh`.

Modification is applied explicitly, so no behavior changes for existing setups.

### DIFF
--- a/smartmon.py
+++ b/smartmon.py
@@ -10,6 +10,7 @@ import sys
 from prometheus_client import CollectorRegistry, Gauge, generate_latest
 
 device_info_re = re.compile(r'^(?P<k>[^:]+?)(?:(?:\sis|):)\s*(?P<v>.*)$')
+COMPAT_DISK_LABELS = False
 
 ata_error_count_re = re.compile(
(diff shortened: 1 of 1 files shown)


dswarbrick:
If you really need this backwards compatible label, you can achieve it with a Prometheus relabel rule.

deajan:
The problem being that both metrics have a disk entry, and only the ones from smartmon.py would need relabeling.
I honestly think it's easier to install new setups so they can be compatible with existing ones (the less work).

deajan:
Closed in favor of #256

PR #250, opened 2025-07-27
Update apt_info.py

code-and-covfefe:
Graceful Dependency Handling
    Comprehensive Error Handling in main
    Structured Entry Point and Exit Codes
    Improved Type Hinting and Code Clarity
    Safer Timestamp Handling
    Edge Case Handling for Package Origins

(Opened new PR because prior one had a typo)

### DIFF
--- a/apt_info.py
+++ b/apt_info.py
@@ -22,98 +22,175 @@ Authors: Kyle Fazzari <kyrofa@ubuntu.com>
          Daniel Swarbrick <dswarbrick@debian.org>
 """
 
-import apt
-import apt_pkg
 import collections
 import os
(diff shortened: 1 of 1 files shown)


code-and-covfefe:
Have just updated the source to fix a subtle bug, so you might want to pull in the updated version unless you want me to open a new PR ?

anarcat (changes requested):
this is a massive change to apt_info.py, with no clear documentation in the commit history on what exactly it does. it conflicts with the current head and likely conflicts with other proposed PRs here, like #234.

could you please break up your work in multiple commits so we can at least try to tell what's going on here?

thanks!

PR #255, opened 2025-09-23
Add nvme-cli 2.3 compat v2

deajan:
Debian 12 (proxmox) ships with nvme-cli 2.3 which has no --verbose parameter for smart-log command.
We check if --verbose exists, and deactivate it if not present.

Also added a change that allows a non error exit when no existing nvme devices are found.

Patch has been tested on Debian 12 and RHEL 9 (with nvme-cli 2.11) in order to ensure there is no regression.


deajan:
Script has since been tested on Debian 13 and RHEL 10 too.

deajan:
Any update on this ? Do I need to make any changes to the PR ?

deajan:
This code has been running in production for 6 months now on multiple RHEL / Debian / Ubuntu machines.
I recently added smart ID 248 from nvme drives in order to be able to calculate write amplification on nvme drives.

Is there anything I can do to get that merged ?

deajan:
Up ;) Still making use of this on Debian and RHEL, both work with said patch.

PR #259, opened 2025-12-14
Add iptables parsing script

TotallyMonica:
# Summary
This PR creates a script that parses iptables and ip6tables rules, collecting the following metrics per rule:

- Table
- Chain
- Rule's line number
- Action (Labeled in `iptables -L` and `ip6tables -L` as target)
- Protocol
- Incoming interface
- Outgoing interface
- Source address
- Destination address
- Extra information
- Rule hit statistics
  - Byte count (as iptables_bytes_total and ip6tables_bytes_total)
  - Packet count (as iptables_packets_total and ip6tables_bytes_total)

# References
PR is largely based on prometheus/node_exporter#1264, implementing the feedback received by it. As such, no files are being written to or read from. 

# Notes
Due to how iptables is designed, iptables must be ran as uid 0, so this means one of the following:
- `/sbin/iptables` must be setuid
- `iptables.py` must be ran as root
- `/sbin/iptables` must have the capabilities `CAP_NET_ADMIN+ep`

# Testing done
- Hosts
  - Ubuntu 24.04.3 Server with iptables v1.8.10
  - Debian 13.2 with iptables v1.8.11


dswarbrick:
Please use the official prometheus_client Python library for rendering the metrics. You can find examples of its usage in other Python scripts in this repo.

We no longer accept Python scripts that implement their own metric rendering functions, since these often miss some of the lesser-known aspects of the Prometheus text exposition format.

TotallyMonica:
I apologize for not seeing this until now, work and holidays lately has made it difficult for me to work on open source projects. I'll get started on the refactor based on the prometheus_client library sometime later this week.

Aside from that, are there any other technical issues present? I have been running this across three VPSes (one Debian 13, two Ubuntu 24.04) across the past four months and have come across no functional issues thus far.

To aide other contributors, it might be worth noting that this library is needed for contributions as it wasn't present 4 months ago and doesn't seem to be present in the README and a CONTRIBUTORS.md doesn't seem to be present.

dswarbrick:
There is a CONTRIBUTING.md that has been under consideration for a while in #146.

Given that you are testing this collector on relatively modern distros (i.e. Debian 13, Ubuntu 24.04), where [iptables is merely a wrapper for nftables](https://salsa.debian.org/pkg-netfilter-team/pkg-iptables/-/blob/master/debian/README.Debian?ref_type=heads), I would ask you to consider whether this collector really makes sense. It is going to be very niche, considering that the majority of systems would have migrated to nftables by now.

Additionally, there are already multiple dedicated exporters for iptables / nftables metrics, e.g. https://github.com/kbknapp/iptables_exporter, https://github.com/metal-stack/nftables-exporter.

TotallyMonica:
I was not aware that iptables on modern distros served as an nftables wrapper, as I was under the impression they were two different options altogether.

As far as the existing exporters is concerned, they didn't show up while I was looking for something along those lines. In that case, this PR can get closed, and I'll move any contributions that I feel are necessary to those projects.

PR #267, opened 2026-05-21
storcli.py: Handle non existent status better

ChristianTackeGSI:
Some controllers (like the SAS9300-8i) do not have a "Controller Status" attribute at all.

Instead of failing completely (and not extracting the other information), just skip the healthy metric.


ChristianTackeGSI:
ping @jcpunk, @LuminatiHD, @dswarbrick: Would anyone please comment on this?

I think, it should be straight forward?

jcpunk:
I've no objection, but I don't have commit rights.

PR #270, opened 2026-08-10
Add name mapping table for smart attrs

deajan:
Some SSD vendors decided that `percent_lifetime_remain` would be called `ssd_life_left`. Let's use a map table in order to fetch those infos in "standard" format

### DIFF
--- a/smartmon.py
+++ b/smartmon.py
@@ -79,6 +79,11 @@ smart_attributes_whitelist = (
     'workload_minutes',
 )
 
+# Name map table for attributes that have different names depending on SSD vendor
+smart_attributes_map = {
+    "ssd_life_left": "percent_lifetime_remain"
+}
+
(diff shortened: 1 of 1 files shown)


dswarbrick:
The human-friendly names are produced by smartmontools, from a lookup table that maps the attribute ID to a label: https://github.com/mirror/smartmontools/blob/master/drivedb.h

It is not the job of this textfile collector to perform yet another layer of translation. You can propose an addition / modification to the upstream smartmontools, or you can override it locally in your smartd.conf.

deajan:
I see... Thanks for the explanation. Indeed we got smart ID 248 and 235 for the same metric, depending on the vendor.
I guess l I'll go the override route, since changing historic settings won't probably be the best idea to suggest to the smartmontools folks.

